PR #10, opened 2023-12-19
Setup and deployment

JonathanHHenson:
- Added additional documentation for running mypy and tests using hatch.
- Coverage now fails under 80%
- Tests action now check coverage and produces a report


hhenson:
Thanks Jon!

PR #22, opened 2024-03-15
Bug fixes and lambdas

abalabin-bamfunds:


src/hgraph/_impl/_runtime/_service_node_impl.py
```diff
@@ -24,7 +24,7 @@ def __init__(self,
         self._active_graph: Graph | None = None
 
     def initialise(self):
-        self._active_graph = self.nested_graph_builder.make_instance(self.owning_graph_id + (self.node_ndx,), self)
+        self._active_graph = self.nested_graph_builder.make_instance(self.node_id + (self.node_ndx,), self)
```
hhenson, line 27: Should we be using the + self.node_ndx now we are using node_id?

src/hgraph/_impl/_runtime/_try_except_node.py
```diff
@@ -44,7 +44,7 @@ def _wire_graph(self):
             node.output = self.output.out
 
     def initialise(self):
-        self._active_graph = self.nested_graph_builder.make_instance(self.owning_graph_id + (self.node_ndx,), self)
+        self._active_graph = self.nested_graph_builder.make_instance(self.node_id + (self.node_ndx,), self)
```
hhenson, line 47: Same question, do we still need the node_ndx?

PR #25, opened 2024-04-09
Initial structure of notebook evaluation logic.

hhenson:


src/hgraph/notebook/_notebook_graph.py
```diff
@@ -0,0 +1,50 @@
...
+    sink_nodes = context.sink_nodes
+    graph_builder = create_graph_builder(sink_nodes)
+    config = GraphConfiguration(start_time=_START_TIME, end_time=_END_TIME)
+    evaluate_graph(graph_builder, config)
+
+
+def notebook_eval_node(self) -> Sequence[tuple[datetime, Any]]:
+    global _START_TIME, _END_TIME
+    record(self)
+    notebook_evaluate_graph()
```
abalabin-bamfunds, line 48: May be in this case the record node should be the only sink being wired?

abalabin-bamfunds (approved):

src/hgraph/notebook/_notebook_graph.py
hhenson: I would like to do that, but need to validate safety of doing so as we could have wired a feedback, this creates sink nodes that MUST be in code to work correctly. So for now any sink nodes that are wired in will still be evaluated as well.

PR #26, opened 2024-04-09
perspective adaptor

abalabin-bamfunds:
includes a simple example with two tables - one editable one with history and a sample layout

### DIFF
--- a/src/hgraph/adaptors/perspective/_perspective.py
+++ b/src/hgraph/adaptors/perspective/_perspective.py
@@ -1,0 +1,348 @@
+import os
+import sys
+import tempfile
+import time
+import concurrent.futures
+from datetime import datetime
+from glob import glob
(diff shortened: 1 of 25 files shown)


src/hgraph/adaptors/perspective/_perspective.py
```diff
@@ -0,0 +1,348 @@
...
+        thread.start()
+
+        thread = Thread(target=_perspective_thread, kwargs=dict(manager=perspective_manager))
+        thread.daemon = True
+        thread.start()
+
+        while not started:
+            time.sleep(0.1)
+
+        print(f"Perspective server started at http://{host}:{port}")
```
hhenson, line 175: Use logging not print for non-debug messaging

hhenson (approved):

PR #38, opened 2024-05-23
Schema extensions, fast TSD and many bug fixes

abalabin-bamfunds:


src/hgraph/nodes/_compound_scalar_operators.py
```diff
@@ -33,7 +33,7 @@ def from_ts_node(**kwargs):
     return from_ts_node(**kwargs)
 
 
-@compute_node(resolvers={COMPOUND_SCALAR: lambda mapping, tsb: mapping[TS_SCHEMA].py_type.scalar_type()})
+@compute_node(all_valid=('tsb',), resolvers={COMPOUND_SCALAR: lambda mapping, tsb: mapping[TS_SCHEMA].py_type.scalar_type()})
```
hhenson, line 36: Is this really required? We only need all_valid if the nodes don't have defaults

src/hgraph/test/_node_printer.py
```diff
@@ -23,7 +23,7 @@ def __init__(self, start: bool = True, eval: bool = True, stop: bool = True, nod
         self.graph = graph
 
     def _print(self, evaluation_clock: "EvaluationClock", msg: str) -> None:
-        print(f"[{datetime.utcnow()}][{evaluation_clock.evaluation_time}] {msg}")
+        print(f"[{datetime.now(UTC)}][{evaluation_clock.evaluation_time}] {msg}")
```
hhenson, line 26: Why the change here?

src/hgraph/nodes/_compound_scalar_operators.py
abalabin-bamfunds: We're producing a scalar from a bundle - the standard data class init requires all non-defaulted fields to be provided otherwise it throws. An alternative would be to generate valid requirements for non-defaulted fields only but that's a bit trickier.

src/hgraph/test/_node_printer.py
abalabin-bamfunds: Python 3.12 deprecates the `utcnow` function

PR #42, opened 2024-06-03
Work on ranking using dependencies

hhenson:
This is an experimental attempt to change how we compute the ranking order in preparation for delayed binding operation.

For review only at this point in time, not ready for merge.

The thing that needs a bit more consideration at this point is the services. They are currently ranked in the `build_services` method.

The idea I was working though was to make each service client dependant on the service_impl using the phantom ranking solution (i.e just add the node_instance to the rank_marker), then make service_impl node instances dependent on the interfaces they make use of, in much the same way as currently we create the dependencies in the current code,  this way we can use the standard ranking code to organise ranking to ensure that a user of a service is always ranked below the implementation and it'd dependencies.

But was having some trouble to work out how to safely obtain the node_instance objects for the clients to do this, obviously the impl's are at handing the build_service logic.


src/hgraph/_wiring/_wiring_node_instance.py
```diff
@@ -80,7 +83,14 @@ class WiringNodeInstance:
     error_handler_registered: bool = False
     trace_back_depth: int = 1  # TODO: decide how to pick this up, probably via the error context?
     capture_values: bool = False
-    _hash: int | None = None
+    _treat_as_source_node: bool = False
+    rank_marker: frozendict[str, "WiringNodeinstance"] = field(default_factory=frozendict)
```
hhenson, line 87: This is the new rank_marker that allows to specify phantom dependencies used for ranking operations only.

src/hgraph/_wiring/_wiring_node_instance.py
```diff
@@ -80,7 +83,14 @@ class WiringNodeInstance:
     error_handler_registered: bool = False
     trace_back_depth: int = 1  # TODO: decide how to pick this up, probably via the error context?
     capture_values: bool = False
-    _hash: int | None = None
+    _treat_as_source_node: bool = False
```
hhenson, line 86: This is used to assist with working around the fact that I have been using compute_nodes for stubbing inputs to inner graphs, these act as source-nodes for the inner graph, so mark them as such to ensure they get ranked correctly.

src/hgraph/nodes/analytics/_recordable_feedback.py
```diff
@@ -72,10 +72,15 @@ def _recorded_source_node(
...
-    return create_wiring_node_instance(node=PythonRecordedSourceNodeWiringNodeClass(signature, object()),
+    from hgraph._wiring._context_wiring import TimeSeriesContextTracker
+    from hgraph._wiring._wiring_node_instance import WiringNodeInstanceContext
+    return create_wiring_node_instance(
+        node=PythonRecordedSourceNodeWiringNodeClass(signature, object()),
                                        resolved_signature=signature,
                                        inputs=frozendict(inputs),
-                                       rank=1)
+                                       rank=1,
+        rank_marker=TimeSeriesContextTracker.instance().rank_marker(WiringNodeInstanceContext.instance())
```
hhenson, line 82: This is the component used to ensure we pick up and context based dependencies.
I may have used this a bit liberally for now, will review which as needed vs superfluous later.

src/hgraph/_wiring/_graph_builder.py
```diff
@@ -52,53 +53,78 @@ def create_graph_builder(sink_nodes: tuple["WiringNodeInstance"], supports_push_
...
+            nodes_to_process.append(to_node)
+            max_rank = max(max_rank, processed_nodes[to_node])
+
+    # Set sink nodes to be max rank
+    for node in nodes:
+        processed_nodes[node] = max_rank
+
+    # Sort nodes by rank
+    result = [node for _, node in sorted((rank, node) for node, rank in processed_nodes.items()) if not node.is_stub]
+    if not all(n.rank <= n_1.rank for n, n_1 in zip(result[:-1], result[1:])):
```
hhenson, line 128: I have this to perform like-for-like comparisons, once the ranking logic is good will remove this and all the current rank properties from the code.

src/hgraph/_wiring/_wiring_node_class/__init__.py
```diff
@@ -1,8 +1,17 @@
 from hgraph._wiring._wiring_node_class._graph_wiring_node_class import *
```
hhenson, line 1: This is just a clean-up of the imports that needed to be done.

src/hgraph/_wiring/_wiring_node_class/_stub_wiring_node_class.py
```diff
@@ -24,3 +24,10 @@ class NonPeeredWiringNodeClass(StubWiringNodeClass):
 
     def __call__(self, _tsb_meta_type: HgTSBTypeMetaData, **kwargs) -> "WiringPort":
         ...
+
```
hhenson, line 27: This was an idea to use a stub node to rank components using the context, but I don't think this will remain as it did not fit well and the phantom inputs (rank_marker) seems like a cleaner approach to the problem.

hhenson:
Merged under a different request

PR #45, opened 2024-06-07
Refactor operators in the library according to the documented table.

syoung-bamfunds:
- Added all the operator overloads for all types where they did not exist in the required packages and added tests for each combination of type and operator
- Renamed all the overloads according to the document
- Added ISSUES.md until we can add to github issues

Breaking changes:
- Moved sum_ into operators as per doc
- _ts_operators no longer exists, as per doc.  Operators in it moved to the correct packages
- Removed (most) overloaded specific operator names from __all__
- and_op renamed to bit_and (and analogies for or_op etc) as per doc
- union_op removed (is equivalent to bit_or)
- match renamed match_ (match is now a python keyword)
- modified_datetime/date removed in favour of Alex's version: hgraph.nodes._time_series_properties.last_modified_time/date

Still to do:
- Add the multi-arg versions of min, max, sum and any others


hhenson:
Will fix issues once merged

PR #48, opened 2024-06-12
Multi arg min max sum

syoung-bamfunds:


src/hgraph/_operators/_operators.py
```diff
@@ -628,50 +628,39 @@ def my_getattr(ts: TS[MyType], attr: str) -> TS[SomeScalar]:
...
+    Binary or multi arg implies item-wise max over all the arguments for collection types,
+    or the maximum scalar value for scalar types
     """
-    raise WiringError(f"operator max_() is not implemented for {lhs.output_type} and {rhs.output_type}")
+    raise WiringError(f"operator max_() is not implemented for {ts.output_type}")
 
 
 @operator
-def sum_(lhs: TIME_SERIES_TYPE, rhs: TIME_SERIES_TYPE = None) -> TIME_SERIES_TYPE_2:
+def sum_(*ts: TSL[TS[SCALAR], SIZE], default_value: TIME_SERIES_TYPE = None) -> TIME_SERIES_TYPE:
```
hhenson, line 655: If we define the ts to be TSL[TS[SCALAR]], ...] then surely the default should be TS[SCALAR] and for output? For the binary operator it is possible to have different lhs and rhs types -> lhs type. But in the general case as defined here?

src/hgraph/_operators/_operators.py
```diff
@@ -703,104 +692,69 @@ def my_len(ts: TS[MyType]) -> TS[int]:
...
-    if len(args) == 1:
-        return args[0]
-
-    return intersection_tsl(TSL.from_ts(*args))
-
-
-@operator
-def intersection_tsl(tsl: TSL[TIME_SERIES_TYPE, SIZE]) -> TIME_SERIES_TYPE:
+def intersection(*tsl: TSL[TIME_SERIES_TYPE, SIZE], tp: Type[TIME_SERIES_TYPE] = AUTO_RESOLVE) -> TIME_SERIES_TYPE:
     """
```
hhenson, line 717: It seems we are combining definition of operator and implementation together, I though we were going to keep the operator as pure interface?

PR #54, opened 2024-07-09
fix a reference issue with free bundles going into inner graphs

abalabin-bamfunds:
and redo frame operators into the standard library APIs


hhenson:
Will fix issues tonight

PR #57, opened 2024-07-11
adaptor wiring proposal

abalabin-bamfunds:


hhenson:
Looks like a good start, I think we should create an example with an actual asynchronous process (e.g. network stack)

src/hgraph/_wiring/_decorators.py
```diff
@@ -551,42 +553,76 @@ def register_service(path: str, implementation, resolution_dict=None, **kwargs):
...
 
     for i in implementation.interfaces:
-        WiringGraphContext.instance().register_service_impl(
-            i, i.full_path(path), implementation, kwargs, resolution_dict
-        )
+        WiringGraphContext.instance().register_service_impl(i, path, implementation, kwargs, resolution_dict)
 
 
-def service_adaptor(interface):
+def adaptor(interface, resolvers: Mapping[TypeVar, Callable] = None):
```
hhenson, line 573: Explain how this different from service interfaces / implementations.
Note what the expected usage pattern is, i.e. what happens if there are multiple instances of adaptor created?

PR #71, opened 2024-08-23
ws adaptors and http trace controller

abalabin-bamfunds:


hhenson:
Will merge and fix failure

PR #99, opened 2024-10-25
precision arg to dedup() needs to be a TS[float]

syoung-bamfunds:
(cherry picked from commit 8430c6ca5adea66e859c490765bbf452dc75af17)

### DIFF
--- a/src/hgraph/_types/_time_series_types.py
+++ b/src/hgraph/_types/_time_series_types.py
@@ -97,10 +97,9 @@ class TimeSeries(ABC):
     @abstractmethod
     def valid(self) -> bool:
         """
-        Is there a valid value associated to this time-series input, or loosely, "has this property
-        ever ticked?". Note that it is possible for the time-series to become invalid after it has been made valid.
-        The invalidation occurs mostly when working with REF values.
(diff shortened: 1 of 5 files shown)


src/hgraph/_types/_time_series_types.py
```diff
@@ -97,10 +97,9 @@ def modified(self) -> bool:
...
-        Is there a valid value associated to this time-series input, or loosely, "has this property
-        ever ticked?". Note that it is possible for the time-series to become invalid after it has been made valid.
-        The invalidation occurs mostly when working with REF values.
-        :return: True if there is a valid value associated with this time-series.
+        Is there a valid value associated to this time-series input, or more generally has this property
+        ever ticked.
+        :return: True if there is a valid value associated to this time-series
         """
 
     @property
```
hhenson, line 105: Seems like a reversion on the docs, or was this intentional?

syoung-bamfunds: No not intentional.  We do see some odd things with git using multiple remotes. My commit didn't include that change in the local repo!  I'll revert locally and include with the next push.  Thanks

PR #100, opened 2024-10-25
improve many things and fix lots of bugs

abalabin-bamfunds:
correct index handling for tables with complex indices, also use wall… 
Fix issue with map inside reduce, by Tope Olukemi
fix signal wiring with free bundles, add some timedelta operators, ad… 
Sub tsd should support any timeseries type, by: Tope Olukemi 
fix an issue with tsd input into map_ being rebound, refactor service… 
Tolerance to dedup() is TS[float] rather than float, by: Simon Young fix tss after rebinding
inspector improvements
fix multiple inheritance of schema types
add a test to check that an empty TSS produces a tick add connect/disconnect logging, heartbeat table and reload UI if_then_else dedup, fix non-peer TSD subscription being too broad conditional formatting and more layout management
upload layouts, promote generic schema to base type add more layout management

### DIFF
--- a/src/hgraph/_impl/_operators/_conversion_operators/_tsb_conversion_operators.py
+++ b/src/hgraph/_impl/_operators/_conversion_operators/_tsb_conversion_operators.py
@@ -14,6 +14,7 @@ from hgraph import (
     DEFAULT,
     OUT,
     TS_SCHEMA_1,
+    HgTypeMetaData,
 )
 
 __all__ = ("convert_tsb_to_bool", "convert_tsb_to_tsd")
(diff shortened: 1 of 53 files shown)


hhenson (approved):

src/hgraph/_impl/_operators/_conversion_operators/_tsb_conversion_operators.py
```diff
@@ -42,6 +43,15 @@ def combine_named_tsb_partial(tp_: Type[TSB[TS_SCHEMA]] = DEFAULT[OUT], **bundle
     return TSB.from_ts(__type__=tp_, **bundle.as_dict())
 
 
+@graph(overloads=convert)
```
hhenson, line 46: Should implement a requirement check to ensure that the schema's are convertible.

PR #158, opened 2025-03-17
rename test dir, fixes and other things

abalabin-bamfunds:
rename tests -> test_hgraph
fix nasty TSS bug
batch TSD queues
add some nodes


src/hgraph/_types/_scalar_type_meta_data.py
```diff
@@ -1079,7 +1079,7 @@ def generic_rank(self) -> dict[type, float]:
 
     def matches(self, tp: "HgTypeMetaData") -> bool:
         return type(tp) is HgCompoundScalarType and (
-            issubclass(tp.py_type, self.py_type) or self.__eq__(tp) or issubclass(self.py_type, tp.py_type)
+            issubclass(tp.py_type, self.py_type) or self.__eq__(tp)
```
hhenson, line 1082: Why was the reverse check removed? Depending on order of calling this can be a problem.

abalabin-bamfunds: Because it allowed incorrect binding - TS[Base] would bind to an input of TS[Derived] so this change was incorrect. The call to `matches` should always be done on the input type and the incoming type passed in otherwise matching compatible things will lose sense.

PR #163, opened 2025-03-17
fix reference issue and suggest an improvement of type inference of values

wukrur:


src/hgraph/_impl/_types/_ref.py
```diff
@@ -230,9 +230,11 @@ def bind_output(self, output: TimeSeriesOutput) -> bool:
 
     def do_bind_output(self, output: TimeSeriesOutput) -> bool:
         if isinstance(output, TimeSeriesReferenceOutput):
+            self._value = None
             return super().do_bind_output(output)
         else:
             self._value = TimeSeriesReference.make(output)
+            self._output = None
```
wukrur, line 237: this was the real issue - rebinding was going from being bound to a reference output to being bound to a value output. `self._output` was not being cleaned hence `value` getter was returning stale value

src/hgraph/_types/_frame_scalar_type_meta_data.py
```diff
@@ -67,11 +68,15 @@ def parse_type(cls, value_tp) -> Optional["HgTypeMetaData"]:
                     raise ParseError(f"Could not parse {value_tp.__args__[0]} as type from {value_tp}")
                 return HgDataFrameScalarTypeMetaData(tp)
 
-        @classmethod
-        def parse_value(cls, value) -> Optional["HgTypeMetaData"]:
+        @class_or_instance_method
```
wukrur, line 71: this might be a little hacky but together with the change in wiring_node_signature gives parsing values a context of the type annotations that is supposed to receive the value. This way bad values can be rejected earlier, good values can get resolved using more context. For example this can potentially solve the empty TSD const, which has been a problem since the beginning. Obviously for it to work well all the types need to be updated, not just this and some pre-resolving introduced before kwargs parsing stage.

PR #230, opened 2026-03-05
New View/Value approach to time-series values

hhenson:
This is the code largely generated by Codex (GPT) with a few helping hands from Claude.
The code is based on the design documentation found in ts_value_v2601. 
The code also includes the first pass at writing C++ nodes.

The code has many issues, not the least being that AI code tends to be hard to read, often full of insane and poor performing aspects. But it does showcase the core value/view concept and provides a solid basis for cleanup and getting the initial structure in place to work from.


hhenson:
OK, This version was made to work, but it is very complicated and messy. I have started on a new version of this with a focus on making the code more supportable. I am closing this request as it does not make sense to merge this code, I am using it instead as a reference to assist with the new port.

PR #385, opened 2026-08-08
Run the test suite on Windows in CI

hhenson:
Windows wheels are published, but nothing in CI ran the suite against them.

`test-matrix` was `runs-on: ubuntu-latest`. Windows built wheels and ran the `-m smoke` subset, and that was all — so a Windows-only failure had been sitting there undetected, because no job existed that could see it.

## Found it by running the suite on Windows

On an x86-64 Windows 10 host, `test_wall_clock_scheduler_reschedule` failed on 5 of 6 legs against `main`, and 8/8 on a focused re-run.

My first instinct was the classic Windows 15.6 ms clock tick. **That is not the cause** — `NtQueryTimerResolution` reports the current resolution at **0.997 ms**. It is scheduling overshoot, not granularity, and raising the timer resolution would not have helped.

Measuring the actual margins over 8 runs on an idle machine:

| | first tick (bounds 50–60 ms) | gap (bounds 90–110 ms) |
|---|---|---|
| macOS | exactly 50.00 | 100.63 – 105.22 |
| Windows | exactly 50.00 | **109.67 – 111.82** |

Only the gap bound is affected, and Windows straddles it — which is why one leg passed while five failed. The first-tick bound has no margin problem on either platform, so it is left alone.

The bound is widened to 130 ms **on Windows only**. Widening it everywhere would have been simpler and worse: the assertion exists to catch scheduler drift, and on Linux and macOS it has ~5 ms of headroom that is worth keeping.

## The CI change

12 legs instead of 6 — ubuntu and windows across 3.12/3.13/3.14 and both runtimes. Each OS now tests the wheel built for it (`release-wheels-${{ runner.os }}`) rather than Linux's. Coverage stays on the single Linux 3.12 leg, so it is measured once.

Linux-only machinery is now guarded: `apt-get`/gdb and core-dump collection run only when `runner.os == 'Linux'`. The venv layout goes through a `VENV_BIN` job env instead of a hard-coded `.venv/bin`, since Windows uses `.venv/Scripts`. The uv cache path is per-OS.

## Verification

Every leg green on the Windows host with this change — 3.12, 3.13, 3.14, both runtimes. macOS unchanged at 1780 / 2457.

I also ran the **published** `hgraph 0.5.39` wheel from PyPI on a clean Windows machine with no source tree: 1777 passed (python) / 2454 (cpp) on both 3.12 and 3.14, with the scheduler test the only genuine failure. So there is no defect in the shipped artifact — the gap was purely one of test coverage.

## Unrelated, noted only

`uv.lock` records `hgraph 0.5.36` while `pyproject.toml` is at `0.5.39`; the release version bump does not appear to update the lock. Left out of this PR.

### DIFF
--- a/.github/workflows/release-wheels.yml
+++ b/.github/workflows/release-wheels.yml
@@ -383,29 +383,37 @@ jobs:
   #
   # Skipped in its entirety on a tag whose commit has already been through it.
   test-matrix:
-    name: Test suite (py${{ matrix.python-version }}, ${{ matrix.runtime }} runtime)
+    name: Test suite (${{ matrix.os }}, py${{ matrix.python-version }}, ${{ matrix.runtime }} runtime)
     permissions:
       contents: read
     needs: [ reuse-build, build-wheels ]
     if: needs.reuse-build.outputs.run-id == ''
-    runs-on: ubuntu-latest
+    runs-on: ${{ matrix.os }}
     # The C++ runtime is not yet at parity with the Python reference implementation, so its legs
     # report without blocking. Every Python version blocks, 3.14 included.
     continue-on-error: ${{ matrix.runtime == 'cpp' }}
     strategy:
       fail-fast: false
       matrix:
+        # Windows wheels are published, so they are tested rather than assumed. Until this was
+        # added the suite only ever ran on Linux, and a Windows-only scheduler failure sat
+        # undetected because nothing here could see it.
+        os: [ ubuntu-latest, windows-latest ]
         python-version: [ '3.12', '3.13', '3.14' ]
         runtime: [ 'python', 'cpp' ]
-        # Coverage is measured once, on the reference leg; the other five would report the same
-        # lines at six times the cost.
+        # Coverage is measured once, on the reference leg; the others would report the same
+        # lines at many times the cost.
         include:
-          - python-version: '3.12'
+          - os: ubuntu-latest
+            python-version: '3.12'
             runtime: 'python'
             coverage_args: '--cov=hgraph --cov-report=xml'
     env:
       HGRAPH_USE_CPP: ${{ matrix.runtime == 'cpp' && 'true' || 'false' }}
       PYTHONFAULTHANDLER: '1'
+      # Layouts differ; every step below runs under bash (Git Bash on Windows) and goes through
+      # this rather than hard-coding one platform's path.
+      VENV_BIN: ${{ matrix.os == 'windows-latest' && '.venv/Scripts' || '.venv/bin' }}
     steps:
       - uses: actions/checkout@v7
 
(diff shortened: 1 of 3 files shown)


hhenson:
Correction to the description: I described the scheduler timings as measured "on an idle machine". They were not.

Checking the host afterwards, it was sustained at 52–95% CPU — Windows Defender (`MsMpEng`, 2391 CPU-seconds accumulated, almost certainly scanning the multi-GB toolchain install and the repeated builds), plus `perfmon` and Task Manager. So the Windows figures of 109.67–111.82 ms were taken on a machine under roughly half load.

This does not change the conclusion, and arguably strengthens it: a shared, throttled CI runner is closer to a loaded machine than to a quiet one, so a bound derived under load is the safer one to ship. But "idle" was wrong and the number should be read as "under moderate load".

Two things still hold independently of load, since both were measured directly rather than inferred: the timer resolution of 0.997 ms (so this is not the 15.6 ms clock tick), and the macOS comparison of 100.63–105.22 ms, which was taken on a quiet machine.

I will re-measure on a genuinely quiet host and post the numbers. If the Windows lag turns out much lower when nothing else is running, the 130 ms bound is more generous than it needs to be and I will tighten it.

hhenson:
All 12 legs green on `64b30f56` — checked job by job, not just the workflow conclusion, since the cpp legs are `continue-on-error` and a red one can hide behind a green tick.

The five Windows failures on the first run were all the **same test**: `test_inspector.py::test_run_inspector`. Every other test passed on every leg (2475 tests, 1 failure each).

It drives the inspector over HTTP and asserts four checks, bounded at 5 seconds. On the GitHub Windows runners the sequence never reached the first check, so `gs.test_state.test_map` raised `AttributeError: test_map` — a message that says nothing about the actual problem.

That bound is a **deadline, not a duration**: `stop_engine(if_true(test.done))` ends the run as soon as the checks pass, so raising it to 30s costs a fast machine nothing and only gives a slow one room. The assertion now reports all four checks by name and distinguishes "never reached" (`None`, the deadline) from "ran and returned the wrong shape" (`False`).

## Correction to my earlier claim

I said all six Windows legs were green on my local Windows host. They were — but **npm was not installed there**, and this test is `skipif(which('npm') is None)`. The box was silently skipping the exact test that fails here. My local sweep showed 131 skips against macOS's 130 and I did not check what the extra skip was.

Node is now installed on that host, so it no longer under-tests: the test runs there and passes 8/8 across both runtimes, in 2–5 seconds. That speed is precisely why it never strained the 5s budget locally while the slower runner could not meet it.

This is the first thing the new Windows legs caught, on their first run — which is the argument for having them.

hhenson:
Re-measured on a genuinely idle machine, as promised when I corrected the "idle machine" claim above. The result contradicts what I expected, so the bound stays as it is.

20 runs per platform, both machines idle (CPU queue length 0 mean and max):

| | min | median | max | bound | headroom |
|---|---|---|---|---|---|
| macOS (arm64) | 100.51 | **105.17** | 105.41 | 110 | 4.6 ms |
| Windows (x86-64) | 100.60 | **111.57** | 111.78 | 130 | 18.2 ms |

Windows gap samples:
```
100.6 100.6 106.6 106.6 107.6 110.7 111.6 111.6 111.6 111.6
111.6 111.6 111.6 111.6 111.6 111.6 111.6 111.6 111.6 111.8
```

**Load was not the cause.** The earlier reading, taken at ~50% CPU, was 109.67–111.82; idle gives 100.60–111.78. Essentially the same distribution. My caveat implied the lag might drop once the machine was quiet and that 130ms could then be tightened — that was wrong.

**It is also not jitter.** Both platforms sit on a tight plateau — macOS ~105.2, Windows ~111.6 — 14 of the 20 Windows samples landing on 111.6 exactly. That is a systematic ~6.4ms offset, not noise. Combined with the timer resolution measuring 0.997ms, neither of the two obvious explanations (coarse clock, busy machine) survives contact with the data.

**So the bound stays at 130.** It has 18ms of headroom over the observed max, against macOS's 4.6ms at 110. Tightening it would make the leg flaky on a 2-core shared runner and buy nothing.

If the ~6ms offset is ever worth pursuing, it is a real difference in how the wall-clock scheduler computes its next wake on Windows rather than a test problem — and it may not survive the engine replacement anyway.

Issue #512, opened 2026-08-19
Implement RFC 0026: Versioned Dataflow Fabric


hhenson:
## Summary

Implement [RFC 0026: Versioned Dataflow Fabric](https://github.com/hhenson/hgraph/blob/main/docs/source/rfc/rfc_0026_versioned_dataflow_fabric.rst) as the C++-first `hgraph-fabric` extension.

The extension distributes recurring hgraph computations through atomic, immutable `Frame` versions. `publish_data` persists a value and publishes a revision notification; `subscribe_data` reconstructs a self-consistent input cut from durable lineage and follows live notifications. Revisions advance when an input-version/output-version tuple changes, including the important case where an input changes but the output Frame does not.

This issue is the implementation plan of record. Each checkpoint should land through a focused, non-draft PR and leave the repository in a validated state.

## Scope

V1 includes:

- immutable full-Frame data versions;
- contiguous acknowledgement revisions with immediate input-version lineage;
- first-writer-wins publication and repairable `as_of`/`latest` indexes;
- independent consistency forests and newest-compatible rolling ancestry;
- live conflated subscriptions, simulation replay, and one-shot snapshots;
- first-class C++ authoring with matching Python exposure; and
- memory, local-filesystem, and S3 persistence plus a Kafka production notifier.

V1 does not include orchestration, a data catalogue, schema evolution under one data id, retention/deletion, delta distribution, event-accurate subscriptions, arbitrary feedback cycles, or multi-data-id transactions.

### `hgraph-persistence` — blocking prerequisite

The current public `FrameStore` contract exposes whole-Frame `write/read/contains/clear`. It does not expose the raw immutable metadata objects, ordered discovery, or conditional reference publication required by RFC 0026. Its current S3 immutable write is an existence check followed by an ordinary output stream, so it is not a cross-process create-if-absent primitive.

Fabric must not implement a private S3/filesystem store. The following reusable persistence-owned contract must land first, aligned with [#498](https://github.com/hhenson/hgraph/issues/498) and RFC 0023's durable-store work:

- [x] Add a public owning, type-erased byte/object-store contract with memory, local, and S3 strategies.
- [x] Support immutable create-if-absent with a typed result distinguishing created, already-identical/idempotent, conflict, and storage failure.
- [x] Support byte-object reads with a typed not-found result rather than conflating absence with transport/authentication failure.
[...]

PR #532, opened 2026-08-20
Add lifecycle-safe asynchronous node wakes

hhenson:
## Summary

- add a public, copyable `AsyncNodeWakeSender` that callback-driven resources can use to request an ordinary node evaluation
- admit callback wakes at the root evaluation boundary, conflate duplicate requests, and wake real-time executors without adding work to graphs that do not opt in
- plan sender storage only for nodes that request the injectable, including nodes inside compiled child graphs
- synchronize sender teardown with concurrent callbacks so post-stop wakes are safe no-ops
- keep wiring-time owned scalar values explicitly owned while preserving the allocation-free borrowed scalar path used by node evaluation

This is the reusable core seam required by hgraph-fabric live subscriptions in #512. It is stacked on #531.

## Validation

- macOS fresh native preset: 1,607/1,607 tests passed
- macOS installed SDK consumer: passed
- Python 3.12 abi3 wheel installed into fresh Python 3.14: 2,123 passed, 9 skipped
- Linux GCC 13 warning-clean build: passed
- Linux native suite: 1,607/1,607 tests passed
- Linux Python 3.14 compatibility suite from abi3 wheel: 2,123 passed, 9 skipped
- ASan callback/shutdown lifecycle tests: passed with leak detection

### DIFF
--- a/docs/source/developer_guide/architecture.rst
+++ b/docs/source/developer_guide/architecture.rst
@@ -246,13 +246,14 @@ The normal cycle shape is:
 
 1. run one-shot before-evaluation callbacks,
 2. notify lifecycle observers before graph evaluation,
-3. scan flattened nodes in rank order,
-4. evaluate nodes whose graph schedule equals ``evaluation_time``, notifying
+3. admit any deduplicated asynchronous node wakes on the evaluation thread,
+4. scan flattened nodes in rank order,
+5. evaluate nodes whose graph schedule equals ``evaluation_time``, notifying
    lifecycle observers immediately before and after each node's evaluation,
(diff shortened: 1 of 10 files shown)


hhenson:
Closing this PR. The `AsyncNodeWakeSender` abstraction is not an acceptable graph ingress path: it lets an external callback address a graph node directly and bypasses the realtime push-source boundary. Realtime ingress must be represented by a root `push_source_node`, whose `PushSourceSender` queues data and signals the executor condition; deterministic simulation work must likewise be generated through a simulation-capable push source. I will rewrite the dependent Fabric subscription/Kafka layers around those existing contracts and retarget the stack directly onto #531.

PR #533, opened 2026-08-20
Implement fabric subscription modes

hhenson:
## Summary

- implement one-shot Snapshot images and half-open Replay from durable as-of indexes
- implement Live subscription-before-image startup, local per-data-id conflation, and slow-client skipping
- retain one newest ahead-of-storage proposal per data id and retry it with bounded backoff until its immutable winning slot validates
- make durable persistence authoritative when a losing proposal payload differs from the accepted slot
- select Auto once during source start from `EngineControlView` and run concrete erased strategies without per-tick mode dispatch
- expose the same operators and runtime behavior through the Python adapter

This completes checkpoint 5 of #512 and is stacked on #532.

## Validation

- macOS native suite: 1,607/1,607 tests passed
- macOS installed SDK consumer, including Fabric probe: passed
- macOS Python 3.14 compatibility suite from Python-3.12-built abi3 wheel: 2,123 passed, 9 skipped
- macOS installed Fabric Python suite: 16/16 passed
- Linux GCC 13 warning-clean build: passed
- Linux native suite: 1,607/1,607 tests passed
- Linux Python 3.14 compatibility suite from Python-3.12-built abi3 wheel: 2,123 passed, 9 skipped
- Linux installed Fabric Python suite: 16/16 passed

PR #534, opened 2026-08-20
Add Kafka-backed fabric notifier

hhenson:
Part of #512.

## Summary

- add the optional installed `hgraph::fabric_kafka` adapter over the public hgraph-kafka service API
- enforce the non-lossy idempotent Fabric profile and correlate publisher delivery before durable revision publication
- reserve ingress before image loading, conflate per data id, validate before explicit offset acknowledgement, and reconcile durable heads on every recovered Live generation
- surface typed lifecycle/failure status and bound outbound, pending-id, receipt, and partition state
- cover fake-broker races/failures, librdkafka protocol recovery, installed SDK consumption, and opt-in external-broker restart conformance

## Validation

- macOS fresh native: 1,607/1,607
- macOS Python 3.14 from Python 3.12 abi3 wheel: 2,122 passed, 10 skipped
- installed C++ `hgraph::fabric_kafka` consumer: 1/1
- external Redpanda broker: round trip passed before and after broker restart
- Linux GCC 13 fresh native: 1,607/1,607
- Linux Python 3.14 with Polars rtcompat: 2,122 passed, 10 skipped
- Linux ASan Fabric suite: 1,862 assertions in 56 cases

Stacked on #533.

### DIFF
--- a/extensions/fabric/CMakeLists.txt
+++ b/extensions/fabric/CMakeLists.txt
@@ -10,6 +10,9 @@ option(HGRAPH_FABRIC_BUILD_PYTHON "Build the optional Python authoring bridge" O
 option(HGRAPH_FABRIC_WARNINGS_AS_ERRORS
     "Treat hgraph-fabric compiler warnings as errors"
     ${HGRAPH_WARNINGS_AS_ERRORS})
+option(HGRAPH_FABRIC_BUILD_KAFKA_NOTIFIER
+    "Build the optional production hgraph-kafka notifier adapter"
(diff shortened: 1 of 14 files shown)


hhenson:
Closing while the underlying subscription-mode layer and this Kafka adapter are rewritten. The current adapter takes a graph-native Kafka subscription into an external Fabric queue and then re-enters the graph through the rejected wake path. The replacement will keep Kafka records on graph edges and will be republished only after the corrected base is ready.

PR #550, opened 2026-08-22
Keep a real-time run alive until end_time

hhenson:
## The defect

A user reported a real-time graph shutting down before its `end_time`, with the implication that it ended once nothing was scheduled. Confirmed: the shared run loop breaks out whenever the graph has nothing scheduled before `end_time`, gated on a predicate that for real time asked only whether the graph contains a push-source node.

```cpp
if (next == MAX_DT || next >= state.end_time)
{
    if (!waits_for_push_sources(state, graph)) { break; }   // realtime: push_source_nodes_end > 0
    next = state.end_time;
}
```

So **any real-time graph without a push source ends the moment its schedule empties**, whatever `end_time` says. Measured against a 2s `end_time`:

| Graph | Requested | Before | After |
|---|---|---|---|
| idle from start | 2.000s | **0.000s** | 2.00s |
| work at +0.5s, then quiet | 2.000s | **0.506s** | 2.00s |

The second is the damaging one — this is not an empty-graph edge case. Any real-time graph dies during its first quiet moment.

It presented as "the real-time engine behaves like the simulation engine" because it literally shared simulation's predicate, where breaking on an empty schedule *is* correct: simulated time is driven entirely by the schedule, so nothing can ever become due.

## The fix

`idle_run_continues` replaces `waits_for_push_sources` — always true for real time, unchanged for simulation. An idle real-time run waits for `end_time`, a producer, or `request_stop`.

`advance_realtime` is reworked around that wait. Each cycle targets `min(next_scheduled_time, end_time)`, waits on the executor condition variable that push producers and `request_stop` already notify, then picks its evaluation time from two rules in strict precedence:

1. **Evaluation time advances** — floor the wall clock at `next_cycle`, since two producer wakes can land in the same microsecond.
2. **Evaluation time never passes `target`** — if the wall clock has slipped past a scheduled cycle, that cycle is still processed at its own logical time (which is also what delivers a late alarm rather than dropping it).

The ceiling must outrank the floor: on the first cycle `target` is `start_time`, already the evaluation time, and that cycle still has to run.

## `max_wait_slice`

New `GraphExecutorBuilder` option, default **100ms**. Both notifications are made under the loop's own mutex and the flags re-read under it, so this bounds a lost wake-up rather than polling. It is also load-bearing for a second reason: an idle graph targets `MAX_ET`, and `MAX_ET - now` in microseconds overflows the conversion `wait_for` performs to the condition variable's clock.

It replaces a hard-coded `TimeDelta{10'000'000}` — microseconds, so a **10 second** slice, not the 10ms its placement suggested.

Also removed: `ready_to_push` (written in two places, never read anywhere) and `last_time_allowed_push`, which only fed the dead 15-second gate around it.

## Why this survived

`real-time NodeScheduler delivers a wall-clock alarm that became due while scheduling` ended with:

```cpp
// The chain never waits for end_time: each due alarm fires next cycle.
CHECK(hgraph::testing::wall_now() - start_time < TimeDelta{4'000'000});
```

Its alarm chain stops rescheduling after 3 evaluations, so it relied on the early exit. The test's actual subject is that the chain progresses promptly rather than stalling, so it now records when the chain completed and asserts on that; `end_time` drops 5s to 1s.

## Doc

`execution_layer.rst` already specified that `end_time` bounds the run and said nothing about exiting early when idle, so this was code drift and the doc was the target. It gains the cycle algorithm, the wait/wake contract, `max_wait_slice`, and the explicit simulation-vs-real-time rule for an empty schedule.

## Verification

- Two new regression tests (idle-from-start; work-then-idle) — fail before, pass after
- Full Catch2 suite on this base: **1591 cases / 255843 assertions, green**
- Full `ctest` (per-test process isolation): **1611/1611** on the same change
- `hgraph.smoke`, `hgraph.header_compile_check`: green

Not covered here: the Python surface was not re-run against a fresh bindings build, so the `run_graph(..., run_mode=REAL_TIME, end_time=...)` path is verified through the C++ executor tests rather than end to end.

### DIFF
--- a/docs/source/developer_guide/data_structures/overview/execution_layer.rst
+++ b/docs/source/developer_guide/data_structures/overview/execution_layer.rst
@@ -40,13 +40,24 @@ Current implementation
     target; or a stop was requested, and the run ends. A lagging graph whose
     wall clock is already past the target does not wait at all.
 
-    Waits are issued in slices of at most
-    ``GraphExecutorBuilder::max_wait_slice`` (default 100ms). Both the push
-    and stop notifications are made under the loop's own mutex and the flags
-    are re-read under it, so the slice is not a poll interval — it bounds the
-    worst case should a notification ever be lost, and keeps the duration
-    handed to ``wait_for`` finite when an idle graph is waiting on a
-    ``MAX_ET`` target.
+    Waits use the pending-push and stop flags as the condition-variable
+    predicate. A producer sets the pending flag under the wait mutex before
+    notifying, and stop follows the same state-before-notify rule. Spurious
+    wakes are therefore absorbed by the condition-variable wait rather than
+    returned to the run loop. Waits are issued in slices of at most
+    ``GraphExecutorBuilder::max_wait_slice`` (default 10 seconds, matching the
+    Python runtime) so an idle ``MAX_ET`` target cannot overflow the duration
+    conversion inside ``wait_for``. A slice timeout before the target only
+    refreshes the wall clock and re-enters the wait; it does not create an
+    evaluation cycle.
+
+    The next evaluation time follows the Python runtime's precedence exactly:
+    ``target = min(next_scheduled_time, end_time)``, followed by
+    ``evaluation_time = min(target, max(previous_evaluation_time + MIN_TD,
+    wall_now))``. Consequently, a producer notification before the target
+    evaluates at wall time (subject to the ``MIN_TD`` monotonic floor), a due
+    target evaluates at its exact logical time even if the wall clock has
+    passed it, and neither case can skip the earliest scheduled cycle.
 
     **Nothing scheduled.** In *simulation* an empty schedule ends the run:
     simulated time is driven entirely by the schedule, so nothing can become
--- a/docs/source/developer_guide/data_structures/overview/execution_layer.rst
+++ b/docs/source/developer_guide/data_structures/overview/execution_layer.rst
@@ -28,6 +28,33 @@ Current implementation
     receive ``evaluation_time`` explicitly, and "next scheduled time" is the
     minimum over the graph's per-node schedule entries.
 
+    **The real-time cycle.** Each iteration takes the minimum of
+    ``next_scheduled_time`` and ``end_time`` as its target. If the wall clock
+    has not reached that target the loop waits on the executor's condition
+    variable — the same one push-source producers notify through
+    ``mark_push_update_pending``, and that ``request_stop`` notifies. It
+    leaves the wait for one of three reasons: the target arrived, so the
+    cycle evaluates at the target (its *logical* time, which is what delivers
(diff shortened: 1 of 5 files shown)


hhenson (commented):
I compared this with the released Python executor (`v_0.5.42`) and exercised the PR through both the native runtime and a Python 3.14 wheel built from this head. The central behavior is consistent: a one-shot real-time graph keeps the run alive after its schedule empties, retains the start-time tick, and treats `end_time` as exclusive. A 120 ms probe took ~125 ms in Python 0.5.42 and ~128 ms through this PR's native Python bridge.

Two follow-ups before merge:

1. **[P2] Avoid the 100 ms periodic wake for indefinitely idle graphs.** `max_wait_slice_{100'000}` makes a default real-time run (`end_time == MAX_ET`) time out and reacquire the executor mutex ten times per second for its entire lifetime, even with no producers or scheduled work. Push and stop both update their predicates under the mutex and notify the condition variable, so responsiveness does not depend on this timeout. The released Python runtime uses a 10-second cap solely to keep the duration finite (`_evaluation_clock.py`, lines 161-166), making the proposed default 100x more active. Please use the Python-compatible 10-second cap (or another non-polling, overflow-safe wait); I would also keep this overflow workaround internal unless callers have a real semantic reason to configure it.

2. **[P2] Add Python-visible regression coverage.** The new behavioral assertions exist only in `tests/cpp/test_realtime_execution.cpp`. The existing Python real-time test (`python/tests/test_signature_compatibility.py::test_realtime_run_defaults_start_time_to_now`) has a 50 ms end time but never asserts that the call remained alive until it, so it passed with the old early-exit behavior. Please add an elapsed/end-time assertion or a dedicated one-shot/empty-schedule Python test so the bridge contract cannot regress independently.

Validation on head `5258c96e`: native `real-time*` tests passed (69 assertions / 18 cases), the focused Python 3.14 wheel probe passed, and `git diff --check` passed.

hhenson:
Addressed both review follow-ups in a13b96a60.

- The real-time wait now uses pending-push/stop as the condition predicate, with state set under the same mutex before notification. Spurious wakes are absorbed, and a forced slice timeout only refreshes wall time and continues waiting. The default finite slice is now the Python-compatible 10 seconds.
- Next-time selection is explicitly min(target, max(previous + MIN_TD, wall_now)), where target is the earlier of the next scheduled time and end time.
- Added a native regression that forces repeated 5 ms slice timeouts yet observes exactly two notified push cycles plus the exact scheduled target cycle.
- Added an installed-wheel Python regression proving an idle one-shot real-time run remains alive for its requested wall-clock window.

Validation: macOS 1619/1619 native and 2128 passed, 10 skipped Python 3.14; Ubuntu GCC 14.2 1619/1619 native and 2128 passed, 10 skipped Python 3.14; Sphinx HTML with warnings as errors; installed SDK consumer.

PR #553, opened 2026-08-24
Use burst ingress for web and Kafka adaptors

hhenson:
## Summary

- use standard burst push sources for Kafka and each independently ordered web channel
- route each Kafka transport burst through one Kafka-specific graph emit whose structural bundle exposes keyed subscription and delivery lanes plus the scalar service-event lane
- retain Kafka transport work in one ordered queue and emit only its collision-free front prefix; the first repeated keyed output, second scalar event, or timestamp/recovery constraint stops dequeueing until the applicable next engine cycle, so no later event overtakes it
- project the Kafka lanes with standard `getattr_` and `map_` composition plus narrow stateless decode nodes; deterministic simulation replay feeds the same emit and projections without introducing a push source
- release web transport admission when a burst enters graph processing; later graph backlog is ordinary graph state, not transport capacity or protocol acknowledgement
- realize burst elements through the active graph value strategy and preserve concurrent compatibility requests by keying before route merges
- update RFCs, user guides, changelogs, topology assertions, and burst/compatibility regressions

## Validation

- macOS fresh native acceptance: 1620/1620 passed
- macOS Python 3.14 stable-ABI wheel compatibility suite: 2128 passed, 10 skipped
- Linux GCC 14 fresh native acceptance: 1620/1620 passed
- Linux GCC 14 Python 3.14 stable-ABI wheel compatibility suite: 2128 passed, 10 skipped
- Windows Visual Studio 2026 / MSVC 19.51 Kafka native target built and service tests passed
- Sphinx HTML build with warnings as errors passed
- `git diff --check` passed

### DIFF
--- a/docs/source/rfc/rfc_0015_kafka_extension_api.rst
+++ b/docs/source/rfc/rfc_0015_kafka_extension_api.rst
@@ -63,20 +63,24 @@ Live Kafka ingress is real-time-only.  One standard unbounded burst push source
 carries discriminated, fully owned transport envelopes for subscription
 values, delivery reports, and service events.  Each sender call admits one
 envelope; one source evaluation transfers all currently pending envelopes as
-an ordered tuple.  The subscription scheduler retains only its required
-timestamp/recovery ordering.  Delivery reports are grouped by a stateless
-classifier and flow through standard ``collect`` and mapped ``emit`` operators;
-the scalar event stream uses standard ``emit`` directly.  This distributes
-independent keys in one public tick, preserves same-key FIFO over later cycles,
-and keeps graph-stop and commit-on-delivery policy on graph.  There is no extension-owned
-cross-thread ingress queue, conflated wake token, or drain node in front of the
-graph.
+an ordered tuple.  One Kafka-specific graph emit classifies that burst into a
+structural service bundle: keyed subscription and delivery lanes plus a scalar
+service-event lane.  Independent keys are emitted together as keyed deltas;
+only a repeated key, a later scalar event, or timestamp/recovery ordering is
+retained for a later engine cycle.  Standard field projection and ``map_`` then
+give each service lane its public shape.  This preserves same-key FIFO without
+head-of-line blocking across independent keys or services, and keeps graph-stop
+and commit-on-delivery policy on graph.  There is no extension-owned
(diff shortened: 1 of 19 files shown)


hhenson:
Pushed `e9dd3fa94` to address the Linux wheel failures.

The exact failed CI core/Kafka wheel artifacts reproduced `ValueView::as_bundle on non-bundle view` in the Kafka burst projections. The projection nodes now retain the erased `ListView` for the duration of evaluation and traverse it by stable indexed access, including the simulation projections, instead of using the range-iterator path. This preserves the O(batch size) hot path and adds no allocations.

Validation completed before push:
- macOS: 1,620/1,620 native tests
- macOS stable-ABI wheel on Python 3.14: 2,128 passed, 10 skipped
- Linux GCC 14: 1,620/1,620 native tests
- Linux stable-ABI wheel on Python 3.14: 2,128 passed, 10 skipped
- Linux GCC 14.2 packaged Kafka wheel: 58/58 tests
- Windows MSVC 19.51: Kafka native target built and service tests passed

The unrelated RFC-28 working-tree changes were not included.

hhenson:
Pushed `7648848d5` with the Kafka ingress redesign. The burst is now classified once by a Kafka-specific graph emit into structural subscription, delivery, and event lanes. Independent keyed work flows in the same engine cycle; only a collision on the same key, scalar FIFO work, or timestamp/recovery ordering is retained and scheduled. Standard `getattr_`/`map_` composition performs the downstream per-service projection, and simulation replay uses the same graph path. Full macOS and GCC 14 Linux native/Python gates pass, as does the MSVC 19.51 Kafka build/test and the warning-as-error Sphinx build.

hhenson:
Updated the Kafka emit design in `44d198d71` to match the clarified ordering model:

- one retained ordered queue, rather than queues per output lane
- dequeue one event at a time into the structural output bundle
- stop at the first event that would write a second value to an occupied single-value output in the current engine cycle
- schedule the next engine time and resume from that same front event; later events cannot overtake it
- retain only per-evaluation output-occupancy scratch alongside the queue

The mixed-burst regression now proves the exact four-cycle boundary progression across a repeated subscription key, repeated delivery id, and second scalar event. Full macOS and GCC 14 native/Python gates pass, as does the Windows MSVC Kafka target.

PR #605, opened 2026-08-27
Extract the one GlobalState publishing protocol

hhenson:
SonarCloud flagged duplication on #569 and it was right — the duplication was mine, and it is still on `main`.

## What it measured

`new_duplicated_lines_density` on #569: **9.11%**, against a **0.4%** project baseline. I queried its duplications API for the specific blocks rather than guessing:

| file | blocks |
|---|---|
| `adaptors/data_catalogue/catalogue.py` | lines 67 and 183 (17–21 lines) |
| `adaptors/data_frame/_data_frame_source.py` | lines 91 and 161 (18–22 lines) |

...plus the cross-file pairs between them. Four copies of the `__enter__`/`__exit__` state-publishing protocol, **character-for-character identical apart from one comment**. I wrote it for `DataStore` and pasted it into `DataConnectionStore`, `DataCatalogue` and `DataEnvironment`. Neither the persistence variant nor the runner was flagged.

## The change

`_published_in_global_state` in `_state.py`, beside the scope guard it uses. A subclass supplies `_STATE_KEY`, and may override `_publish` / `_unpublish` when it does more than store itself there — the hook that would let the persistence storage adopt it later rather than becoming a fifth copy.

**+58 / −96 lines.**

## Why extract rather than tolerate

The protocol is fiddly in exactly the places copies drift: restoring a *shadowed* entry rather than deleting, closing only a scope **this** block opened, and holding both on the error path. Four copies is four places for those to diverge silently.

Concretely, the extracted version already fixes something **none of the four copies did**: it closes the scope if publishing raises. That bug existed four times and would have had to be found and fixed four times.

## Verification

- Every adaptor suite that exercises these classes: **96 passed, 1 skipped**.
- Full core suite: the same 10 pre-existing subprocess-harness failures as before the change, **nothing new**.
- Checked the importers: `hgraph_persistence` uses `_active_global_state` and `_global_state_scope`, neither of which moved or changed signature.

**Not yet run:** the six-suite reproduction on `hg-linux` — the box went unreachable partway through (`Connection refused`). This is a Python-only refactor whose own tests run locally, so the local signal is meaningful here in a way it was not for the persistence failures; but I would rather say that plainly than imply full coverage. I will run it when the box is back and report on this PR.

### DIFF
--- a/python/hgraph/_wiring/_state.py
+++ b/python/hgraph/_wiring/_state.py
@@ -281,6 +281,56 @@ class _global_state_scope:
         return False
 
 
+class _published_in_global_state:
+    """Mixin for a ``with`` block that publishes itself into the GlobalState.
+
+    Four adaptor stores share one protocol, and it is fiddly enough in the
+    corners -- restoring a shadowed entry, closing only a scope this block
+    opened, holding both on the error path -- that four copies of it is four
+    places for those corners to drift apart.
+
+    A subclass supplies ``_STATE_KEY`` and may override ``_publish`` /
+    ``_unpublish`` when it has more to do than store itself under that key.
+    """
+
+    _MISSING = object()
+
+    def _publish(self, state):
+        state[self._STATE_KEY] = self
+
+    def _unpublish(self, state):
+        if self._previous is self._MISSING:
+            state.pop(self._STATE_KEY, None)
+        else:
+            state[self._STATE_KEY] = self._previous
+
+    def __enter__(self):
+        # The block publishes into the GlobalState for the wiring inside it to
+        # find, so it needs a state to publish into: open one when the caller
+        # has not, and close it in __exit__ -- the state must not outlive the
+        # `with` that needed it.
+        self._state_scope = _global_state_scope()
+        state = self._state_scope.__enter__()
+        try:
+            self._previous = state.get(self._STATE_KEY, self._MISSING)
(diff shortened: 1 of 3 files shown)


hhenson:
Full six-suite reproduction done on `hg-linux` — the leg this PR listed as unrun. Core wheel plus all five extensions built from source against the installed SDK, exactly as `release-platform-wheel.yml` does it, with every CI dependency present.

| suite | result |
|---|---|
| analytics | **57 passed** |
| fabric | **16 passed** |
| kafka | **58 passed** |
| persistence | **82 passed** |
| web | **75 passed** |
| core | 2223 passed, **6 failed** |

The 6 core failures are all `python/tests/adaptors/sql/test_sql_adaptor.py`, and they are **pre-existing**: the identical six fail on `origin/main` built the same way (`6 failed, 6 passed` there), for want of the optional `connectorx` driver, which neither environment has.

That baseline matters here rather than being a formality — two of the six (`test_catalogue_publish_routes_to_sql_write_adaptor`, `test_catalogue_subscribe_routes_batch_sql_source`) exercise the very `DataCatalogue` / `DataEnvironment` classes this PR refactors. Same names, same failures, both sides.

One note on the earlier run: analytics failed to build in a first attempt and built cleanly here with identical flags. That was two heavy jobs of mine competing on the box, not a defect.

PR #631, opened 2026-09-01
Prototype experimental hgraph language and design guides

hhenson:
## Summary

- add `language/` as an independently buildable, opt-in project that consumes the public hgraph SDK without becoming a core dependency
- implement the first `hgl` frontend and direct-wiring slice used by `check`, `test`, `run`, and `repl`
- keep the prototype deliberately free to make incompatible syntax, AST, semantic, and lowering changes while the language model settles
- keep design, user, developer, example, compatibility, and roadmap documentation with the language project

## Structural type prototype

- parse and print nominal `struct`, `abstract struct`, public `export struct`, generic applications, trailing `requires`, `null`, named construction, and contextual `delta<S>(...)`
- resolve nominal struct identity, effective inherited fields, defaults, optionality, generic argument roles, constraint names, reflection functions, and operator requirements
- permit inheritance only from abstract structs; concrete structs are implicitly final, field type and optionality remain invariant, and a child may replace an inherited default
- evaluate statically decidable requirements such as closed type-set membership before admitting an explicit generic struct specialization
- lower scalar/defaulted/inherited structs to hgraph Bundle values and type-only generic structs to invariant Bundle specializations
- lower simple field-wise temporal struct construction to public TSB structural wiring, preserve `atomic<S>` harness values, and construct sparse scalar deltas without applying defaults
- keep nominal operator identity in the shared hgraph registry and resolver rather than adding a second HGL-only overload dispatcher

## Existing language model retained

- use `fn` for both composition and runtime functions, with node-only statements determining runtime classification
- use explicit `impl fn` for operator candidates and bodyless nominal `operator` contracts
- use canonical value types plus recursive temporal interpretation, with `atomic<T>` as the explicit scalar boundary
- use ordinary `return` and injected `out`; legacy `emit` remains rejected
- distinguish immutable `let`, evaluation-local mutable `var`, persistent recordable `state`, lifecycle blocks, output access, and collection traversal
- keep push adaptors, transports, callbacks, threads, and arbitrary native resources in C++ extension modules

## Explicit prototype boundaries

The implementation fails closed, with tested diagnostics, for unsettled or missing native contracts: multiple-parent field order, constructor inference, typed `const` generic Bundle metadata, explicit optional-field clearing, temporal structured deltas, nested temporal construction requiring scalar expansion, runtime-function lowering, callable generic substitution, and source-defined operator candidates.

Operator requirements are represented and nominally bound in the frontend, but overload ranking and final admission remain the responsibility of hgraph's shared `TypePattern`/`ResolutionMap`/`requires_` machinery.

## Validation

- clean warnings-as-errors native build with `HGRAPH_BUILD_LANGUAGE=ON`
- `ctest --preset cpp --parallel 2`: 1,667/1,667 passed
- stable-ABI wheel built with Python 3.12 and installed into a fresh Python 3.14 environment
- `pytest python/tests -q -m "not wip"`: 2,223 passed, 10 skipped
- focused language CTest suite: 12/12 passed
- all relative Markdown links under `language/` resolve
- `git diff --check`


hhenson (commented):
## Review

**Verdict:** the scaffold is clean and the design records are unusually disciplined about not growing a second runtime — one resolver, one registry, public SDK only, classification from source not from codegen. I'd merge the scaffold and docs as the design preview they claim to be, after fixing the doc-internal contradictions below (all doc-only). Five design points are worth deciding *before* Slice 1 starts, because each one shapes the grammar or the core-SDK ask. One process point.

### What I verified

- **Standalone build against an installed SDK** (`cmake -S language -DCMAKE_PREFIX_PATH=<wheel site-packages>`): builds with `-Werror`, both CTest smoke tests pass, `hgl --version` → `hgl 0.1.0 (hgraph 0.8.0)`. It needed `-DPython_EXECUTABLE=<venv python>`; without it `hgraphConfig.cmake` sets `hgraph_FOUND=FALSE` ("Python-enabled SDK requires nanobind"). The README's standalone recipe omits that.
- **Opt-in repository build** (`-DHGRAPH_BUILD_LANGUAGE=ON -DHGRAPH_WARNINGS_AS_ERRORS=ON`): `hgl` target builds, `ctest -R hgraph_language` 2/2. Default build untouched (guarded `add_subdirectory`), sdist exclusion present.
- **All 60-odd relative Markdown links and anchors resolve.**
- **Every hgraph API claim in the docs cross-checked against the tree** — all accurate: `TSW<T, Period, MinPeriod = 0>` + `TSWAny` wildcard only (no named size variables), `TSL<T, 0>` is dynamic, `ts_input/list_view.h` exposes only `modified_*` ranges, `OperatorRegistry` has keyed installers + `run_installers()` + `reset()` but no provider-scoped removal, `RecordableState<TSB<...>>::field<"">()`, `InputValidity::Unchecked`, `last_modified_time()`, `keys_`, `add_`/`div_`, `if_then_else`, `map_`, `TSDDataView::key_set()`. The `rolling<f64, 20>` → `TSW<Float, 20, 20>` lowering is right: 0.5's `TSW[float, Size[20]]` normalises min to 20, whereas the C++ static default is `MinPeriod = 0` (valid from the first tick) — so the generator must always emit the third argument, which the doc does.
- Good news for the nominal-identity design: analytics already registers as `Operator<"hgraph.analytics.rolling_mean", ...>`, so `(module, name)` → dotted registry key is an existing convention, not a new one.

### Doc-internal contradictions (fix before merge)

1. **`accumulator` silently drops `b`** (`user-guide/functions.md` § State). Both `when` blocks end in `return total`, and § Ordered activation says a `return` in an earlier block prevents later blocks from running. When `a` and `b` tick in the same cycle, `total -= b` never executes and that modification is lost forever (`modified(b)` is false next tick). The `combined_total` form with `out = total` is the correct one; `accumulator` should use it, and the text should warn that `return` inside an ordered handler is a way to lose ticks.
2. **`if` on a temporal condition in a composition body.** `normalize_price` (`functions.md` § Public functions) is `if value > 0.0 { value } else { 0.0 }` where `value` is temporal. `examples/runtime-choice.hgl` says wiring-time `if` is for `const` only and temporal selection is `if_then_else(...)`, and the grammar defines no if-expression. Pick one: either `if` over a temporal condition auto-lowers to `if_then_else` (then say so and drop "deliberately different"), or it's a `phase` diagnostic (then fix the example).
3. **Python ternary in the README.** `out = (out if valid(out) else 0.0) + value` (`user-guide/README.md`) — that spelling exists nowhere in the grammar; `functions.md` writes the same function with `if/else` statements.
4. **`list<T>` has no fixed-size spelling.** The iteration tables distinguish "fixed TSL" (`modified` only) from "unbounded TSL" (`added/modified/removed`), but the type grammar can only produce `list<T>` → `TSL<T, 0>` (dynamic). `rolling<T, N>` has a size; `list<T, N>` should too, mirroring `TSL[T, Size[N]]`.
5. **Parameter shadows an intrinsic.** `first_modified_index(values: list<f64>)` names its parameter after the `values` traversal intrinsic. The prelude names (`keys`, `values`, `items`, `valid`, `modified`, `added`, `removed`, `delta`) aren't reserved and no shadowing rule is stated — the example demonstrates the hazard rather than the rule.
6. **`start`, `stop`, `state` as hard keywords** will collide with ordinary parameter names (`fn window(start: datetime, stop: datetime)`). They are only meaningful at function-body top level; make them contextual like `out`.
7. **`valid`/`modified` are phase-polymorphic but only the runtime half is written down.** `key_set` gets an explicit "composition → live projection, runtime → borrowed view" rule. `valid(x)`/`modified(x)` in a composition body presumably wire std `valid`/`modified` (TS<bool>) — state it, otherwise `if valid(x) { ... }` in a composition body is undefined.
8. **No `duration` scalar.** `last_modified(v) > some_time` needs datetime comparison, alarms/scheduler need durations, and duration windows are on the roadmap. RFC 0002 already defines Instant/Duration/etc. — say which one `datetime` is and add `duration`.

### Design decisions I'd make before Slice 1

**A. Make implementation binding explicit.** Today a same-named `fn` becomes a public, globally-registered operator candidate purely because a selective import happens to bring an operator of that name into scope. Adding `use market.pricing::{value}` for an unrelated call site silently promotes a private `fn value` helper into the target closure's candidate universe — and can produce ambiguity errors in *other* packages. The docs cite Rust traits and Swift protocols, but both require an explicit conformance declaration. An `impl fn value(...)` (or `fn value(...) for market.pricing::value`) removes the surprise and makes three special rules unnecessary: "`export fn` on an operator-bound fn is invalid", "a later import cannot reinterpret an exact function", and the per-short-name uniqueness rule on imports.

**B. Recordable-by-default state is the wrong default.** The std library has 189 `State<>` uses against 22 `RecordableState<>`, and its stated criterion (`arithmetic_impl.h:422`) is *loopback* state only. `RecordableState` is a TSB output per node with per-field modified tracking — precisely the overhead the "fused node" pitch is selling the removal of. Suggest plain state by default with an explicit `record state` (or keep the default and write down why it beats hgraph's own practice).

**C. Registration removal, leases and unloading are scope creep into core — and avoidable.** Their only consumer is REPL module replacement (Slice 3), yet the roadmap puts "provider-scoped candidate provenance, installer removal, registration removal, and live-plan lease support" in Slice 2, and `modules.md`/`compiler-and-lowering.md` spend roughly a third of their length on it. `hgl run` already executes in a child process and the docs already permit the REPL to rebuild the whole session. If REPL replacement rebuilds in a fresh child, none of provider removal, installer unregistration, leases, or image unloading is needed. Slice 2 then needs only what exists today: init + keyed replayable installers. Move the removal machinery to Slice 4, conditional on in-process replacement proving necessary. (CLAUDE.md §3(i) — don't build machinery the current milestone doesn't need.)

**D. Non-atomic heterogeneous tuples have no target schema.** `tuple<f64, f64>` → "structural tuple of temporal children", but hgraph has homogeneous `TSL` and named `TSB` only. The docs flag it open; it needs closing before Slice 1 because it's in the type grammar. Options, cheapest first: require `atomic<tuple<...>>` for heterogeneous tuples in slice 1; lower homogeneous `tuple<T, T>` to `TSL<T, 2>`; or map to the existing `UnNamedTSB` with positional fields.

**E. The hgraph-side asks need an RFC in `docs/source/rfc/`.** The design requires four core public-API changes: TSW named max/min size patterns in `ResolutionMap`, `added`/`removed` ranges on the public TSL input view, the module-descriptor format, and (if C survives) provider registration handles. Per RFC 0000 those are "cross-project capabilities owned by hgraph" and today they live only in `language/docs`, which is designed to leave the repo. One RFC (next is 0031) listing exactly what the language needs from core, with the language docs pointing at it, keeps the ask governed where core changes are.

### Minor

- `hgl --version` prints `hgraph 0.8.0`: that's the CMake *native API* version, not the release (0.8.2x). Label it (`SDK API 0.8.0`) or it will read as a stale install.
- `hgraphConfig.cmake` requiring nanobind for a pure-C++ consumer is a core-side wart this PR surfaces; worth an issue.
- `hgl --help --version` exits 2 — harmless.
- `use hgraph.std::{map}`: fine, but the kernel descriptor should say which std names drop the trailing underscore (`map_`, `add_`, `keys_` → `map`, `+`, `key_set`).
- The scalar table should say how RFC 0003 registered scalars surface (presumably via extension descriptors).

hhenson:
Pushed fdcfba0 applying the decisions on the review above (docs and examples only, 16 files):

- **A — `impl fn`.** Grammar, reserved words, binding rules, diagnostics, lowering, test matrix, and every example/guide now use `impl fn` as the only way to implement an operator. Plain `fn` beside an in-scope operator name is a `name` conflict; `impl fn` with no operator in scope is a `module` error; `export impl` is invalid. Justification recorded in `syntax-and-semantics.md` § Generics and nominal operator binding.
- **B — state stays recordable-by-default.** `language-model.md` / `syntax-and-semantics.md` now say why: `state` is by definition a time series; cached handles and similar belong to a separate resource concept still to be designed.
- **C — lifecycle kept as designed.** `modules.md` § Generated module lifecycle now opens with the rationale (Python exposure already relies on load/unload, overload registration must be evaluated for hand-written C++ modules too, deterministic teardown for shutdown and in-process unit tests).
- **D — tuples.** `tuple<T0, T1, …>` → hgraph's un-named bundle with positional fields `_0`, `_1`, …; literal index = field projection; a homogeneous tuple is deliberately not a `list<T, 2>`. Removed from the open-questions lists.
- **E — RFC promotion** noted in the roadmap under the Slice 2 hgraph-side requirements.
- **List sizes.** `list<T>` unbounded, `list<T, n>` fixed, `unbounded` as the sentinel a `const` size generic may bind. Lowers to `TSL<T, SIZE<"n">>` / `TSL<T, 0>` — hgraph's `TSL` pattern already binds a size variable to the concrete size (0 for dynamic), so unlike TSW this needs no core change. `array<T, n>` recorded as considered-and-rejected.
- Also fixed contradictions 1, 3, 5 (accumulator `return`, README ternary, `values` shadowing).

Still open from the review, each needs a decision: (2) `if` on a temporal condition in a composition body — auto-lower to `if_then_else` or diagnostic; (6) `start`/`stop`/`state` hard vs contextual keywords; (7) composition-phase meaning of `valid`/`modified`; (8) a `duration` scalar.

hhenson:
Pushed `90e761af7`: a **proposal** for the temporal scalar types (review item 8), marked "proposed" in `syntax-and-semantics.md` § Temporal scalar types.

**Types** — the four zone-free RFC 0002 core types, all hard reserved words:

| HGL | hgraph | Value |
| --- | --- | --- |
| `date` | `CivilDate` (`Date`) | proleptic Gregorian, 0001-01-01..9999-12-31 |
| `time` | `CivilTime` (`Time`) | µs since midnight, `< 24:00:00` |
| `datetime` | `Instant` (`DateTime`) | UTC instant, µs — the engine clock / `last_modified` type; **not** a civil date-time |
| `duration` | `Duration` (`TimeDelta`) | signed µs; a day is 86 400 s, no months |

`civil_datetime`, `period`, `zone_id`, `zoned_datetime`, ranges stay library scalars under their registered names.

**Literals** — single tokens, validated and normalized in the lexer:

```hgl
@2026-09-03                 // date
@09:30:00  @09:30:00.250    // time
@2026-09-03T09:30:00Z       // datetime (an offset like +01:00 is accepted and normalized to UTC)
5m  1.5h  -250ms  1h + 30m  // duration: one unit per token (d h m s ms us); composites via +
```

Rejected at lex time: `@2026-02-29`, `@24:00:00`, `@2026-09-03T09:30:00` (no offset — naive-means-UTC never enters the language), `0.5us`, `1h30m`, `5min`. Canonical printed spelling: `@YYYY-MM-DD`, `@HH:MM:SS[.ffffff]`, `@…Z` always UTC, durations in the largest exactly-dividing unit (`90m`, `1500ms`, `0s`).

**Arithmetic** — hgraph's table, identical in composition, runtime and constant folding: `datetime ± duration`, `datetime - datetime → duration`, `date ± duration` (whole-day part, Python parity), `date - date → duration`, duration `+ - neg * / ÷number`, `duration / duration → f64`, same-type comparison only. Everything else (`datetime + datetime`, `time ± duration`, `date + time`, cross-type compare, `rolling<T, 5m>`) is a `type` diagnostic.

**hgraph-side asks** (added to the roadmap list): reversed `i64|f64 * duration` overloads and `Time` comparison overloads — neither is registered in std today.

Alternatives noted in the spec: bare ISO tokens (clash with `-` and `:`), `date"…"` prefixes, constructor calls.

hhenson:
Decisions applied in `762301743`: `@` sigil and reserved type names agreed; shorthand accepted where unambiguous (`@09:30`, `@2026-09-03T09:30Z`, `+01` offsets, multi-part durations `1h30m` / `2m30.5s` in descending unit order); canonical spelling is the shortest form (`@09:30`, `1h30m`, `1s500ms`); duration scaling is duration-first only (`cooldown * 2`; `2 * cooldown` is a `type` diagnostic), so the reversed-order std overload ask is dropped. The `Time` comparison overload ask remains.

hhenson:
Pushed `20e3c6f44`: **proposed** zoned temporal types (named zones for date-times and for times of day), extending § Temporal scalar types.

| Type | hgraph | Literal |
| --- | --- | --- |
| `civil_datetime` | `CivilDateTime` | `@2026-09-03T10:30` (the offset-free shape — a distinct type, never an instant read as UTC) |
| `timezone` | `ZoneId` | `@[Europe/London]` |
| `zoned_datetime` | `ZonedDateTime` | `@2026-09-03T10:30+01[Europe/London]` — RFC 9557; offset required |
| `zoned_time` | `ZonedTime` **(new hgraph type)** | `@09:30[America/New_York]` |

Rules: a `zoned_datetime` literal always carries its offset (that is how the two `01:30`s of a fold day are told apart, and a skipped time cannot be written); `@…T10:30[Europe/London]` without an offset is rejected with a hint to `resolve(civil, zone, ambiguous:, nonexistent:)` — a literal never picks a fold/gap policy silently. Zone names are validated syntactically at compile time; existence and offset consistency are checked by the run's provider (strict decoding), so a literal's meaning never depends on the compiler's TZDB. Zoned arithmetic never constant-folds.

Arithmetic (hgraph's overloads): `civil_datetime ± duration`, `civil − civil → duration`, `date + time → civil_datetime`, `zoned_datetime ± duration` (timeline; offset re-resolved), `date + zoned_time → zoned_datetime` under the `Reject` policies (hgraph's precedent: `+` with a period uses `MonthEndPolicy::Reject`), with `resolve(date, zoned_time, …)` for explicit policies. Zoned values are `==`-comparable structurally only; ordering is a `type` diagnostic (`to_instant`/`to_civil`/`same_instant`), as RFC 0002 requires.

hgraph-side asks (roadmap): `ZonedTime` core scalar + its operators/codecs/Python wrapper as an RFC 0002 amendment; ordering overloads for `Time` and `CivilDateTime` (`register_ordered_same_scalar_comparisons` covers neither).

Naming: `timezone` rather than `zone` so `zone` stays usable as a parameter name (hgraph's own zone operators name their parameter `zone`). Bracketed `@[Europe/London]` rather than bare `@Europe/London` because `+`, `-`, `/` would otherwise become token characters — can be relaxed later, not withdrawn.

hhenson:
Pushed `1f7a77a0d`: duration rolling windows, and the zoned types recorded as agreed.

**Rolling windows** (`syntax-and-semantics.md` § Types, user guide § Rolling windows, lowering, roadmap, testing):

- `rolling<T, max[, min]>` sizes are `i64` tick counts *or* `duration` spans; the argument type selects the kind, both arguments must be one kind (`rolling<f64, 5m, 3>` is a `type` diagnostic).
  ```hgl
  rolling<f64, 20>          // the last 20 values, valid once it holds 20
  rolling<f64, 20, 5>       // the last 20 values, valid once it holds 5
  rolling<f64, 5m>          // the last five minutes, valid once it spans 5m
  rolling<f64, 5m, 1m>      // the last five minutes, valid once it spans 1m
  ```
- Omitted minimum = maximum for both kinds (0.5's `TSW[T, WindowSize[timedelta]]` normalises the same way). Tick sizes positive; duration maximum positive; duration minimum may be `0s` — the only way to say "valid from the first value" for a duration window, since a one-value window has span zero.
- Semantics are hgraph's, as `plans_and_ops/time_series.rst` records them: a tick window keeps the newest `max` values; a duration window keeps every value within `max` of evaluation time (no element bound — memory follows tick rate, said explicitly). Validity is minimum-gated for both; the duration minimum is measured over the captured span, not elapsed run time (the C++ runtime's rule; 0.5 used run time — the language follows the runtime).
- Identity: kind + resolved sizes. `rolling<f64, 5m>` ≡ `rolling<f64, 300s>`; `rolling<f64, 20>` ≢ `rolling<f64, 20s>`; kinds never unify. A `const` size generic has its declared type (`i64` or `duration`), so one generic declaration accepts one kind.
- Lowering: tick → static `TSW<T, max, min>`; duration → `TypeRegistry::tsw_duration(value_type, time_range, min_time_range)` (the same interned schema Python's `TSW[T, timedelta]` produces). `static_schema.h` has no duration `TSW`, `TypePattern::tsw` matches no duration window, and `tsw_any` is the only wildcard — so the roadmap ask now reads: a compile-time duration `TSW` marker, a concrete duration type pattern, and named size variables of either kind.
- Still open (moved to the open-questions list): rolling-window iteration (both kinds share the window view: `values`, `time_values`, `value_times`, `removed_value`), and a parameter spelling that accepts either kind (`TSWAny`). Nothing is permitted for the latter yet.

**Zoned types → agreed.** Status line, the `ZonedTime (proposed)` cells, and the `resolve(date, zoned_time, …)` wording now read as the record, with a note that a revision replaces the text. Example added: `summarize_recent(window: rolling<f64, 5m>)`.

hhenson:
## Tests, running, and how the first compiler pass will execute programs (`96dd11ca9`)

Three proposals, each written into the docs as the record; veto by editing or replying.

### 1. Tests: `test` / `assert` / `eval` (spec § Tests and the evaluation harness)

```hgl
test midpoint_waits_for_both_sides {
    assert eval(midpoint, tob: [(1.0, _), (_, 3.0), (2.0, _)]) == [_, 2.0, 2.5]
}
```

- `test name { ... }` is a module-scope declaration; it sees unexported functions (tests live beside the code), never lowers into an artifact, and `hgl test` runs it.
- `eval(callee, args...)` is **syntax**, not a function — the arguments are typed by the callee. `const` params take constants; temporal params take a sequence. `assert` takes any wiring-time `bool`.
- **Dense sequences** are one element per engine cycle and `_` is "no tick" (input) / "did not tick" (output). This is exactly hgraph's `eval_node` alignment (`MIN_ST + i*MIN_TD`, padded to the longer of inputs/output), so a test here means the same as the Python/C++ harness test.
- **Timed sequences** `[0s: 1.0, 2m: 3.0]` (offset from start) or `[@2026-09-03T08:00Z: 1.0, ...]` (absolute, fixes the start) seed the sparse absolute-time replay and record sparsely; the expected side lists exactly the ticks produced. All-dense or all-timed per `eval`.
- `test assert eval` are hard reserved; a lone `_` is the placeholder token (identifiers may still start with `_`). Sequence literal outside a test = constant `list`; tuple literal = constant tuple (one-element needs a trailing comma).
- Open: explicit end bound, approximate float comparison, delta spellings for set/map/list elements.

### 2. Running: no `main`, no `run(fn, config)` (spec § Running a module)

An **entry** is any `export fn` with no temporal parameters. A run binds it to a mode, clock, and `const` values from *outside* the source, so the same module runs unchanged as a backtest and as a live process:

```text
hgl run app.hgl [--entry name] [--mode sim|realtime] [--start <datetime>]
        [--end <datetime|duration>] [--set name=<const expr>]... [--config run.toml]
```

```toml
[run]
entry = "main_graph"
mode = "realtime"
start = 2026-09-03T08:00:00Z
end = "1d"

[run.params]
window = 20
symbols = ["AAPL", "MSFT"]
```

CLI overrides the file; TOML types map onto `i64`/`f64`/`str`/`bool`/`datetime`/`date`/`time`/`civil_datetime`/`list`, and a string binds any temporal parameter through the HGL literal spelling. Defaults are hgraph's `run_graph` defaults. Output is `time value` lines. I recommended against `run(fn, config)` in-source because it hard-codes the deployment into the module; the config-file/CLI split keeps that outside.

### 3. Execution: a direct-wiring backend (architecture § Two backends, one wiring — **direction-level, please read**)

To "experiment for real" without waiting for C++ codegen, the first pass executes composition-only programs by walking the typed IR and calling hgraph's **public erased dispatch** — `wire_operator(Wiring&, name, span<WiringArg>, output_required, expected_output)` in `operator_dispatch.h`, the same entry the Python bridge uses — plus hgraph's own `replay`/`record` harness nodes, and then running the graph in process on the unmodified runtime. Nothing in the language project runs at tick time, so this is not an interpreter of hgraph behaviour (the non-goal stands); it is wiring-time interpretation that delegates every wiring step to hgraph, and it builds the identical graph the generated `wire<Op>()` calls would.

- `hgl test`, `hgl run`, `hgl repl` use it for composition-only programs (no native toolchain needed).
- Runtime functions (`state`/`when`/...) have no wiring-time form: they fail with a `backend` diagnostic naming the function until the C++ backend lands (Slice 2). No slow-path emulation.
- Both backends must record the same ticks; the parity suite runs the test corpus through each, C++ is the reference.

Roadmap Slice 1 is now "atomic frontend and direct wiring" and ends with `hgl check`/`test`/`run`/`repl` working on composition-only programs — which is the acceptance for "parser-only is not a milestone".

### Next

Starting the compiler first pass in `language/src/` on this branch: lexer (with the temporal-literal validation), AST + parser, diagnostics, `hgl check` with an AST/IR dump, then the resolver/type checker against the hgraph registries and the direct-wiring `eval`/`test`/`run`/`repl` engine. Tests go into `language/tests/`.

hhenson:
## Compiler first pass: the syntax layer is in

`language/src/` now contains the frontend the developer guide describes
(`compiler-and-lowering.md` "Frontend components", `syntax-and-semantics.md`),
and `hgl check` runs it:

- `src/syntax/`: `source` (file + line table, half-open byte ranges),
  `diagnostic` (category-tagged sink; renders `path:line:col: category:
  message` + source line + carets, ordered by position), `temporal` (the
  full literal grammar and validation of the "Literals" section, canonical
  re-spelling, exact duration arithmetic), `token`/`lexer` (one `Newline` per
  run of terminators, comments as trivia, `_` placeholder, `@` literals lexed
  by shape so `@2026-09-03-1d` is `date - duration`, `1h30m` one token,
  `5min` an unknown-unit diagnostic), `ast` (index-based arena, `std::variant`
  payloads, no pointers), `parser` (recursive descent + precedence climbing,
  one diagnostic per bad construct, recovery at the points the spec lists),
  `ast_printer` (`--dump-ast`).
- `src/driver/`: `hgl check <file> [--dump-tokens] [--dump-ast]`, exit
  0/1/2; `test`/`run`/`emit-cpp`/`build`/`repl` are reserved and say so.
- Tests: Catch2 `hgl_syntax_tests` (118 cases, 1787 assertions: temporal,
  lexer, parser) plus one CTest case per `examples/*.hgl` running
  `hgl check`. All six examples parse with zero diagnostics, including the
  new `test` block in `midpoint.hgl`. Tree green under `-Werror` in both the
  standalone SDK build and the `-DHGRAPH_BUILD_LANGUAGE=ON` repository build.

### Grammar decisions recorded in the spec (implementation forced them)

All in `syntax-and-semantics.md`; shout if any should go the other way:

- **Newlines.** Ignored inside `()`/`[]`/`<>`, after `,`, after a binary or
  assignment operator, `=`, `=>`, `->`, and a type/named-argument `:`, and
  before a line starting with a binary operator, `->`, or `else`. `!` never
  continues; postfix `(`/`[`/`.` never cross a newline; `{` sits on the
  header's line; two statements on one line is a diagnostic; `;` is
  diagnosed as "not a statement terminator".
- **Statement/expression productions** are now written out (they were only
  a precedence table). `if cond block [else (block | if)]` is a *primary*
  expression usable as a statement, so `if … {} else {} + 1` needs
  parentheses. Spec item 2 (temporal conditions) stays open at the
  classifier, as agreed: the parser accepts `if` uniformly.
- **Types.** `tuple/list/set/map/rolling/atomic` are container types only
  when directly followed by `<`; size expressions parse at additive
  precedence; `unbounded` is the sentinel only directly before `>`.
- **`in`** is contextual (identifier elsewhere); it is not in the reserved
  list and I kept it that way.
- **Imports** are always braced: `use a.b::x` is a diagnostic
  (`use a.b::{x}`); the grammar already said so, the parser now agrees.
- **Temporal:** an instant's UTC value must stay within years 0001–9999
  (so every instant has a canonical spelling); `@0001-01-01T00:00+23:59`
  is rejected.

### Doc updates

Status notes (`README.md`, `docs/README.md`, guide READMEs,
`modules-and-tools.md`, `examples/README.md`) now say what `hgl` does;
`compiler-and-lowering.md` describes the implemented syntax layer;
`testing-and-compatibility.md` names the Catch2 suites; the standalone build
recipe notes `-DPython_EXECUTABLE`. Root `CMakeLists.txt` adds `language`
after `tests` so the language suites reuse the repository's Catch2.

### Next

Resolver + type checker against hgraph's registries, then the direct-wiring
backend (`src/wiring/`) so `hgl test`, `hgl run`, and a first `hgl repl`
execute composition-only programs for real (roadmap Slice 1).

hhenson:
## Slice 1 first pass: resolver, direct-wiring backend, `hgl test` / `run` / `repl`

Commit `cd0b376f1` (rebased over the generics/struct doc commits — no overlap). The language can now be tried for real:

```sh
hgl check examples/midpoint.hgl        # parse + resolve
hgl test  examples/midpoint.hgl        # midpoint_ticks ... ok
hgl run   file.hgl --entry ticker --set every=1500ms --end 3s
hgl repl
```

### What landed

- **`src/semantics/resolve`** — name resolution and function classification (`compose` / `runtime` / `lifecycle`) over the AST, per the syntax guide's new *Scopes and name lookup* section (block locals → params/generics → module decls → selective imports → prelude intrinsics). Kernel binding follows the interim table in `compiler-and-lowering.md`: `hgraph.std::x` → registry `x` else `x_`, `hgraph.analytics::x` → `hgraph.analytics.x`; any other module is a `module` diagnostic until descriptors exist.
- **`src/wiring/backend`** — the direct-wiring backend. Composition-only functions wire through `wire_operator` against the live `OperatorRegistry`; `test` / `assert` / `eval` run on the replay/record harness with dense sequences (`_` = no tick, output padded to the longest input, `==` is equal length + `Value::equals`); `run_program` wires an `export fn` whose parameters are all `const` (or `--entry`) into a scratch unit that prints `time value` per tick, in sim or realtime.
- **`hgl` driver** — `check` now resolves; `test [names...]`, `run [--entry] [--mode] [--start] [--end dt|duration] [--set name=expr]`, and `repl` (declarations accumulate into a session module; other input runs as a synthetic test with retained `let`/`var` bindings; bracket-depth continuation; `:list`, `:help`, `:quit`).
- **Tests** — `tests/semantics/resolve_tests.cpp`, `tests/wiring/backend_tests.cpp`, plus a CTest case that runs `hgl test examples/midpoint.hgl`. 12/12 `hgraph_language` CTest cases green in both the repository build (`-DHGRAPH_BUILD_LANGUAGE=ON`) and the standalone build against the installed wheel SDK, `-Werror` on.
- **Build note** — executables linking the backend against the Python-enabled SDK need `hgraph_link_python_embedding`; `hgl_link_runtime_executable()` applies it when the SDK provides it (no-op in the repository build).

### First-pass limits (recorded in `testing-and-running.md#first-pass-limits` and `compiler-and-lowering.md` "First pass")

- Composition functions only; `state` / `when` / lifecycle bodies resolve and classify but are a `backend` diagnostic when wired ("'counter' is a runtime function").
- Dense sequences only; a timed sequence is a `test` diagnostic.
- Harness parameters are atomic / scalar; heterogeneous `tuple<f64, f64>` needs `atomic<...>` (open item D).
- `hgl run` without `--config`; entry = `export fn` with all-`const` params, overridable via `--set`.
- Diagnostics still print hgraph schema names (`float`, `timedelta`, `Tuple[float,float]`), and a sim run over a `schedule` source with no `--end` runs to `MAX_ET`.
- `requires` clauses and `struct` declarations from the two newer doc commits are not parsed yet (the docs say so).

### Untouched review items

Conditional syntax (2), the shadowing rule (5, the guide records what the compiler does meanwhile), non-temporal state (B), and RFC promotion of the hgraph-side asks (E) are still open. Next language steps in order: timed sequences, structural tuples in the harness, TOML `--config`, then the C++ backend.

PR #656, opened 2026-09-04
Add wiring-scale benchmark workloads

hhenson:
**Stack 5 of N**, based on #655 → #654 → #651 → #648. Retarget as parents merge.

The benchmark pack measured wiring time through one wide/deep graph only, so the 148-second portfolio wiring fixed in #636 had no number attached until it was profiled by hand. Three one-cycle workloads whose measured time is wiring + build, all with `independent_size` so `--size-scale` grows them:

| id | shape | modes |
|---|---|---|
| `construct_higher_order_py` | 40 `map_` layers, each over a *distinct* Python child graph with a nested `switch_` and its own lambdas | all (0.5 comparison possible) |
| `construct_dispatch_cases` | a `dispatch` with 24 registered leaf cases called from 40 sites | C++-first only |
| `construct_overloads` | an operator with 32 node overloads resolved at 200 call sites | C++-first only |

Distinct callables per layer matter: with one shared child the first cut wired in 4 ms because the wiring caches collapsed it, which is not what an application with hundreds of graphs looks like.

**Performance round (second commit, `benchmarks/results/*-stack-20260904-macos.*`).** Stack tip against the pinned 0.8.19, M4 Max, five fresh-process samples:

| area | result |
|---|---|
| wiring | wide/deep 1.9x faster, layered higher-order 3.0x, dispatch cases 2.7x, overloads 1.3x (that is #636 landing on top of the 0.8.19 pin) |
| runtime | flat on 70 of 80 workloads including both switch scenarios and mesh; the switch per-tick rebind removal in #655 is within noise |
| memory | flat against `main`; +0.1 to 0.2 MB on tick loops and +3% on large retained capacity versus 0.8.19, present on `main` too |

**Attribution.** A first cross-run comparison suggested the map-only dense TSD workload was 17% slower than `main`. An interleaved A/B (a `main` wheel and the stack wheel in separate venvs, `benchmarks/runner.py` alternated for nine rounds per scenario, same machine state) put the stack within one percent of `main` on every dense TSD workload; `main`'s own samples are bimodal on this machine (63.6 to 76.7 ms), so single-run comparisons of separate wheels carry roughly ten percent of noise. Conclusion: **no runtime regression from this stack**. The dense TSD map/reduce (+9 to 12%) and JSON round-trip (+22 to 25%) gaps against 0.8.19 are present on `main` and predate the stack; they deserve the same interleaved treatment against a 0.8.19 wheel before anyone acts on them.

Guard test added (deterministic totals); `benchmarks/README.md` records the workloads.


benchmarks/results/performance-stack-20260904-macos.md
```diff
@@ -0,0 +1,184 @@
...
+- fixed release wheel: hgraph-0.8.19-cp312-abi3-macosx_15_0_arm64.whl
+- fixed release SHA-256: e7c4f19920a45ce9da0d4e4c479af2fd258e4f55b0f2de0215e5b105548629d1
+- current-source compiler: Apple clang version 21.0.0 (clang-2100.1.1.101)
+- current-source revision: 6e47d30b393d
+- current-source fingerprint: 915e4b4a8728f66f6d50b347beb46a15ec56104da5700070e80edc13dcb7b69a
+- current-source build type: Release
+- cycle scale: 1.0
+- size scale: 1.0
+- fresh-process samples: 5
+- modes: hgraph 0.8.19 (`release`), current source (`hg-cpp`)
```
hhenson: Current code is not hg_cap but a version (or current of hgraph)

benchmarks/scenarios.py
```diff
@@ -360,6 +360,142 @@ def g():
...
+
+        _register()
+
+    family = (base, leaves, describe, score)
+    _EVENT_FAMILIES[cases] = family
+    return family
+
+
+def _dispatch_sites(cases, sites):
+    base, leaves, describe, _score = _event_family(cases)
```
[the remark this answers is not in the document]
hhenson: Fixed in d8ef83bc2. Dispatch and overload families are now built by the scenario factories before timing and split so each workload registers only what it measures. A regression test verifies the family builders are not called during graph execution. Corrected five-sample measurements are in benchmarks/results/wiring-scale-20260905-macos.md.

benchmarks/results/performance-stack-20260904-macos.md
hhenson: Updated. Reports now call this current hgraph; hg-cpp remains only the internal command-line mode identifier. The historical report and corrected wiring-scale report both use the new user-facing label.

hhenson: Follow-up fixed in a3a8b7684. The internal benchmark mode is now current throughout the CLI, environment naming, validator, tests, and result metadata. Obsolete hg_cpp distribution fallbacks were also removed; version discovery uses the hgraph package.

PR #663, opened 2026-09-05
Define Python-object value semantics once in the bridge

hhenson:
**Stack 2, PR 5 of N**, based on #662 → #660 → #659 → #658. Retarget as parents merge.

Family 6 of the retrospective (Python-object handling inside the type layer), first step. Three translation units each decided what hashing and equality mean for a stored `PyObject`: the Python-owned Bundle entry (`bridge_state.cpp`), the retained-value entry (`types/metadata/value_plan_factory.cpp`) and the bridge's `PyObj` hash (`value_conversion.cpp`). #555 added the address-hash fallback and the equal-values-keep-equal-hashes rule to two of them separately, and the retained entry rendered strings through a different path.

`python_bridge::object_hash` / `object_equals` / `object_compare` / `object_str` are now the one definition; all three sites delegate. Behaviour is unchanged except that the retained entry renders through the same UTF-8 path as the bundle entry. `PyObject_Hash` is in exactly one translation unit: ratchet `python-object-hash-units` 3 → 1. Verified: Python-enabled build links; `python/tests` 3251 passed.

Not in this PR, and recorded in the design record as the remaining layering step for that file: moving the retained entry itself out of `types/metadata` behind a bridge-registered provider (the type layer still carries 13 Python conditionals in that file).

Design record: `python_bridge.rst`, "Value and reference crossings".

### DIFF
--- a/include/hgraph/python/bridge_state.h
+++ b/include/hgraph/python/bridge_state.h
@@ -3,6 +3,7 @@
 
 #if HGRAPH_ENABLE_PYTHON_USER_NODES
 
+#include <compare>
 #include <hgraph/hgraph_export.h>
 
 #include <nanobind/nanobind.h>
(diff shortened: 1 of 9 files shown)


include/hgraph/python/bridge_state.h
```diff
@@ -54,6 +55,23 @@ struct HGRAPH_LOCAL PythonValueHolder {
...
+ * equal so equal values keep equal hashes; ordering maps a Python comparison
+ * error to ``unordered`` and an empty side below a live one. Each acquires
+ * the GIL itself. ``object_hash`` requires a live object; callers decide
+ * what an empty holder means.
+ */
+[[nodiscard]] HGRAPH_EXPORT std::size_t object_hash(PyObject *object);
+[[nodiscard]] HGRAPH_EXPORT bool object_equals(PyObject *lhs, PyObject *rhs);
+[[nodiscard]] HGRAPH_EXPORT std::partial_ordering object_compare(PyObject *lhs,
+                                                                PyObject *rhs) noexcept;
+[[nodiscard]] HGRAPH_EXPORT std::string object_str(PyObject *object);
```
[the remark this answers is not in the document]
hhenson: Addressed in 75f2eb0: the contract now lives in `include/hgraph/python/object_semantics.h` and the CPython implementation in `src/hgraph/python/impl/object_semantics.cpp`; `bridge_state.h` no longer declares them, and the retained-value entry and the PyObj hash include the contract header rather than the bridge-state facade. The Bundle entry's diagnostic type name goes through `object_str`, which retires bridge_state.cpp's private UTF-8 helper. python_bridge.rst module table and the ratchet owner updated.

src/hgraph/python/impl/object_semantics.cpp
```diff
@@ -0,0 +1,111 @@
...
+      return std::partial_ordering::less;
+    }
+    const int greater = PyObject_RichCompareBool(lhs, rhs, Py_GT);
+    if (greater < 0) {
+      PyErr_Clear();
+      return std::partial_ordering::unordered;
+    }
+    return greater == 1 ? std::partial_ordering::greater
+                        : std::partial_ordering::equivalent;
+  } catch (...) {
```
hhenson: Scope.h

hhenson: Converted in 1cd57b02e on this PR, so the new file never carries a hand-rolled boundary: `object_compare` uses the three-argument `fallback_on_exception(unordered, f, on_error)` with `PyErr_Clear` as the on_error step. #668 (stacked here) converts the remaining 23 sites and adds the `catch-all-sites` ratchet.

PR #683, opened 2026-09-05
RFC 0034: C++-first NATS and TSD cache extension

hhenson:
## Summary

- propose a separately installed, C++-first hgraph-nats extension using the official NATS C client
- define distinct Core NATS, JetStream, and Key/Value service contracts over one graph-scoped resource
- add typed publish_tsd and subscribe_tsd composition for exposing and consuming TSD state through NATS KV
- define full-child images, generation/head commits, bounded per-key conflation, freshness heartbeats, publisher fencing, and re-image recovery
- document that KV replication is state-convergent but does not preserve atomic multi-key source ticks
- connect the broker-specific design back to the layered network-services research

## Validation

- /tmp/hgraph-nats-rfc-docs-venv/bin/python -m sphinx -E -W --keep-going -b html docs/source /tmp/hgraph-nats-rfc-docs-html
- git diff --check origin/main...HEAD


hhenson (commented):
This is acceptable as a first pass draft.

PR #701, opened 2026-09-05
Parity: carrier templates for the RFC 0033 fix shapes

hhenson:
## Summary

RFC 0033's last open commitment: the parity catalogue gains templates for the four type-argument fix shapes the 2026-09 fix-series retrospective catalogued, each written in its released 0.5 spelling and each making the materialised type observable on every tick, so a wrong carrier form diverges rather than merely wiring.

| Template | Shape |
|---|---|
| `type_argument_positional` | the type as the positional argument after the value: `const(value, TS[float])`, `nothing(TS[int])` under `default()` |
| `type_argument_default_order` | a bare graph subscript fills the `DEFAULT` variable ahead of an `AUTO_RESOLVE` one declared first |
| `type_argument_size_pin` | `type[SIZE]` resolved from a `TSL` input or pinned by `g[Size[n]]` reaches the body as the `Size` object |
| `type_argument_collection` | a `type[SCALAR]` carrier resolved from `TS[tuple[int, ...]]` / `TS[frozenset[int]]` reads back as the parameterised generic |

Seven `coverage-type-argument-*` corpus recipes pin them (corpus-only, like the other public-contract templates). Every one replays identically against the local 0.5 reference environment.

Two reference differences surfaced on the way and are recorded as accepted deviations in `parity_matrix.rst`:

- 0.5 raises `TypeError` at runtime for `const(3, TS[float])`; the candidate converts the value at the selected type. The recipe passes a float, since its subject is the positional carrier.
- 0.5 spells a frozenset carrier `collections.abc.Set[int]`; the candidate hands back `frozenset[int]` (the reverse binding returns what was written). The template compares the generic structurally.

Validators bound every parameter; controller tests cover them; `parity_testing.rst` lists the templates; the RFC's Parity paragraph records the commitment as met.

## Test plan

- [x] `python -m tools.parity validate`: 101 recipes.
- [x] `python -m tools.parity replay` of all seven recipes: reference and candidate both `ok`, difference `null`.
- [x] `test_parity_harness.py` + `test_all_valid_parity.py`: 62 passed.

### DIFF
--- a/docs/source/developer_guide/parity_matrix.rst
+++ b/docs/source/developer_guide/parity_matrix.rst
@@ -581,6 +581,17 @@ Types and scalars
    * - ``TSL``
      - Full
      - Fixed and dynamic.
+   * - Type arguments (``type[...]``, RFC 0033)
+     - Full, two accepted deviations
+     - Matched, deferred, materialised and ranked by the registry; the parity
+       templates ``type_argument_*`` replay the four fix shapes identically
+       against 0.5. Accepted deviations found by them: ``const(1, TS[float])``
+       converts the value at the selected type here, where 0.5 raises
+       ``TypeError`` at runtime (additive); a ``type[SCALAR]`` carrier resolved
+       from ``TS[frozenset[int]]`` reads back as ``frozenset[int]`` here (the
+       reverse binding hands back what was written) where 0.5 spells it
+       ``collections.abc.Set[int]`` -- the same parameterised generic, which
+       is what the template compares.
    * - ``Frame[Rows, Metadata]``
      - Full for typed value transport and metadata-aware table transforms
      - Metadata is encoded field by field in Arrow schema metadata: supported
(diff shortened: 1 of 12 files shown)


hhenson:
Review pass on the merged head:

- **Codex** (code + security review on 0cee865): completed with no findings, so nothing to answer there.
- **SonarCloud** "8 New issues": all eight are `python:S5778` on `python/tests/test_parity_harness.py` — each `pytest.raises` block in `test_type_argument_templates_reject_unbounded_parameters` held both the recipe construction and the `validate_recipe` call, so a `RecipeError` raised while *building* the recipe would have satisfied the assertion. Fixed in 6b6ad6f: the accepted and rejected cases are parametrised tables, the recipe is built outside the block and only `validate_recipe` runs inside it. Same eight message assertions, no change to `tools/parity`.

The merge of main (retained-entry provider, #698) into this branch is what CI is now running; the full local gate (native rebuild, `uv sync` reinstall of every native member, all Python suites including the adaptor and extension suites, consumer checks) is running on 6b6ad6f and I will post the result here.

hhenson:
Local gate on 6b6ad6f (main merged in, Sonar fix): native `_hgraph` rebuilt, `uv sync` reinstall of every native member, then core + adaptor + extension Python suites: 3606 passed, 10 skipped; C++ unit tests, the installed extension consumer, the install consumer and the fabric test package all pass. SonarCloud re-analysed the new head with 0 new issues. CI on this head is in progress.

PR #730, opened 2026-09-06
Type layer: container conversions move to the bridge (RFC 0035, PR 2 of 5)

hhenson:
Second implementation step of RFC 0035 (#720, merged). Follows #722.

**What moves.** The compact list / tuple / shaped-array / cyclic-buffer / queue / set / map / key-set-adapter conversions and the mutable list / map / set read-backs leave `compact_container_ops.h`, `compact_container_ops.cpp` and `mutable_container_ops.h` for `src/hgraph/python/impl/container_conversions.cpp`. The ops tables hold forwarders into the new `PythonOps::Compact` and `PythonOps::Mutable` sections; the list variant (list / tuple / ndarray) is still selected at binding time, now between three forwarders. The bodies read only the public storage API (`compact_storage.h`, `mutable_container_ops.h`) and rebuild through the value builders, so no private detail header was needed; the two headers lose their guarded `conversion.h` includes.

**Plan adjustment.** The TSD proxy surfaces first planned for this step read the proxy's private context and belong with the TS data families in PR 4; the RFC's implementation plan records the renumbering (98 → 72 → 14 → 0).

**Ratchets**: `type-layer-python-conditionals` 120 → 98; `type-layer-nanobind` 525 → 426.

**Tests**: `test_python_ops.cpp` pins that a compact slot is a provider forwarder selected by family (named error without a provider, resolved after registration) and that the read-only key-set adapter keeps its null `from_python` slot.

**Gate** (macOS): core-only tree builds (the bodies are gone from the type layer, nothing conditional remains in these files); native `_hgraph` rebuilt; `uv sync` reinstall of every native member; core + adaptor + extension suites 3591 passed, 10 skipped, plus the ratchet and registry-snapshot tests; core C++ suite 1702 cases; the three consumer checks; `sphinx -W`. Linux GCC 14 and Windows MSVC validations are running; results will be posted here.


hhenson:
Linux validation of cd8e3ff: Ubuntu x86_64 / GCC 14.3, CI core-only configuration with warnings as errors — build exit 0, `ctest` 1708/1708; Python-enabled build of hgraph and every native member under gcc-14 — 3609 passed, 9 skipped; installed extension consumer and install consumer pass. Windows/MSVC is running on the same commit.

hhenson:
Windows validation of cd8e3ff: Windows 11 x86_64 / MSVC 14.51 — native build with warnings as errors: exit 0, `ctest` 1708/1708 on the freshly built binary; MSVC build of the `hgraph` wheel and the persistence and web native members via `uv sync`: `python/tests` 3294 passed, 9 skipped; installed Python extension consumer check passed. cd8e3ff is green on macOS (local gate), Linux/GCC 14 and Windows/MSVC.

PR #759, opened 2026-09-06
RFC 0036 PR 2: std operators call the owners of REF transparency (stdlib-ref-dereference 67 -> 0)

hhenson:
Implements RFC 0036, PR 2 of 4 (`docs/source/rfc/rfc_0036_reference_transparency_owners.rst`, "Implementation plan"): the std operators stop dereferencing for themselves and call the owners PR 1 (#758) introduced. Stacked on #758; merge order #758 → this.

**What changes**

The 67 `dereference(` sites under `include/hgraph/lib/std` and `src/hgraph/lib/std` map onto the five intents the RFC catalogued:

- *Own bound argument* (27): `time_series_schema_at(context, i)` / `time_series_schema(arg)` where a call context or `WiringArg` is at hand; `NamedPort::observed()` in a `compose`; a `VarIn` element is left as supplied, because `value_argument` already observed it for the parameter.
- *Collection element* (9): `TypeRegistry::value_element_ts(collection)`, including the four `ref(dereference(element_ts))` spellings, which become `ref(value_element_ts(collection))`.
- *Comparison through references* (17): `time_series_value_equivalent(a, b)`; `paired-dereference-comparisons` 12 → 1.
- *Key sources and raw `compose` ports* (7): the argument helper gains a **port** form, `time_series_schema(const WiringPortRef &, mode)`, for wiring code that holds ports rather than a call context (the switch key, dispatch slots, mesh keys, `tsb_itemwise`'s inputs); the bridge's target-inference inputs (`convert_target.cpp`) use its **schema** form. Both live beside the existing argument form in `operator_type_resolution.h`.
- The structural hop in `resolve_indexed_reference_target` is `TSOutputView::through_reference()`.

**Findings** (RFC implementation status)

1. `tsb_ref_field_node` (behind `getitem_` / `getattr_` on `REF[TSB]`) and `dereference_tsb_ref` / `dereference_tsl_ref` dereferenced their declared schema in `eval`: a registry lock every time the reference ticked. They now observe the referenced container once in `start` (`RefContainerState`), and `tests/cpp/test_ref_projection_locks.cpp` pins it: a reference retargeting every cycle through both projections costs no type-system lock per tick.
2. The switch terminal check compared the branch's dereferenced output against the raw switch output; it now uses `time_series_value_equivalent`, so interior references on the switch-output side are transparent too, as they already are at the binding that follows.
3. The `to_data_frame` start hook and target resolver dereferenced schemas binding had already observed; the no-ops are gone.

**Ratchets**

[...]

PR #760, opened 2026-09-06
RFC 0036 PR 3: the runtime's structural hops and shared-output stability call the owners (runtime-ref-kind-probes 7 -> 0)

hhenson:
Implements RFC 0036, PR 3 of 4 (`docs/source/rfc/rfc_0036_reference_transparency_owners.rst`, "Implementation plan"): the runtime's seven `TSTypeKind::REF` probes become calls to the owners. Stacked on #759; merge order #758 → #759 → this.

**What changes**

- *Structural hops* (`output_at_path` in `graph.cpp`; `walk_ts_path` and the key-set case of `walk_source_to_output` in `nested_bindings.h`): `view = view.through_reference()` replaces the probe-then-`dereference`-then-`binding_for` sequence; a non-reference is returned as is.
- *Forwarding tree root-REF adaptation* (`bind_forwarding_output_tree_to_source`): asks whether the source *refers to* the target's shape, `time_series_schema_equivalent(source.schema()->referenced_ts(), target.schema())`, which is false for a non-reference, so the kind probe in front of it was redundant. Interior REF fields stay visible to the tree as before.
- *`make_race_tsd_node`*: the output is `registry.ref(element)`, idempotent for a dict that already holds references. **Finding**: the output was declared as a reference to the *dereferenced* element; it is now `REF[element]`. The difference is interior references only, which the reference resolves at its target and consumers compare through `time_series_value_equivalent`.
- *Shared-output stability* (`input_target_is_stable`): `input.is_bindable() && input.bound() && !input.bound_target_is_reference()`, the record the link wrote when it bound (PR 1), replacing the per-tick schema-kind probe and the `target_link_storage` check.

**Tests**: `test_shared_output_node.cpp` gains "shared output capture follows a source reference that retargets every cycle": the capture's input is bound to a `REF[TS[int]]` output that alternates between two constants; the capture stays active and a consumer of the shared output sees 101, 202, 101 across three cycles. (The mock harness has no wiring-time same-cycle pairing, so the test orders the capture before its source and ranks the consumer after the capture, as the wiring proof would.)

**Ratchet**: `runtime-ref-kind-probes` 7 → 0. **Docs**: `testing.rst` bullet, RFC 0036 implementation status (PR 3 + finding).

**Validation (macOS, local gate)**

- C++ core-only suite: all tests passed (257436 assertions in 1715 test cases)
- Python gate (`python/tests` incl. adaptors + all extension suites): 3622 passed, 10 skipped
- `sphinx -W`: clean
- Consumer checks (`python_extension_consumer`, `install_consumer`, fabric test package): passed
- Architecture ratchets: 18 passed with `runtime-ref-kind-probes` at 0

Linux (GCC 14, warnings-as-errors) and Windows (MSVC) results follow as comments.

### DIFF
--- a/docs/source/developer_guide/testing.rst
+++ b/docs/source/developer_guide/testing.rst
@@ -208,7 +208,9 @@ Each entry names the layer that owns the rule:
   ``time_series_value_equivalent`` (``endpoint_schema.h``) owns
   reference-transparent equivalence (RFC 0036; the count is that owner);
 * the runtime probing ``TSTypeKind::REF`` per tick, when a node's REF handling
-  mode is fixed when the node is built;
+  mode is fixed when the node is built (held at zero since RFC 0036: a
+  structural hop goes through ``TSOutputView::through_reference()`` and the
+  shared-output capture reads the record its link wrote at bind);
 * Python wiring choosing a type carrier by operator name, or keeping a shadow
   schema-to-Python-type dictionary, when the resolver and the registry own
   both (the dictionaries are gone since RFC 0033's PR C: the bridge's
--- a/docs/source/rfc/rfc_0036_reference_transparency_owners.rst
+++ b/docs/source/rfc/rfc_0036_reference_transparency_owners.rst
@@ -505,7 +505,23 @@ Implementation status
   start hook and target resolver dereferenced schemas that binding had
   already observed (a value input's schema, the ``time_series_schema_at``
   argument); the dereferences were no-ops and are gone.
-* PR 3 (runtime), PR 4 (Python wiring): pending.
+* **PR 3 (runtime)** -- landed: ``runtime-ref-kind-probes`` 7 → 0. The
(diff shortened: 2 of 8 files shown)


hhenson:
Linux validation of 8d8483d7f, the merged head (Ubuntu x86_64 / GCC 14.3, CI core-only flags with warnings-as-errors, then the Python build of every native member; checkout hash confirmed in the log):

- C++: build clean, ctest 1721/1721 passed
- Python gate: 3623 passed, 9 skipped
- installed Python extension consumer: passed

hhenson:
Windows validation of 8d8483d7f, the merged head (hg-windows128, Win11/AMD, MSVC static, CI core-only flags; checkout hash confirmed in the log):

- C++: build clean, ctest 1721/1721 passed
- Python core suite (`python/tests`): 3308 passed, 9 skipped
- installed Python extension consumer: passed

macOS, Linux and Windows are all green on this head.

PR #796, opened 2026-09-09
docs(hgspec): prototype backend-neutral runtime contracts

hhenson:
## Summary

- split the exploratory `.hgspec` runtime-contract work from the mergeable HGL language changes
- describe backend-neutral value, time-series, execution, operator, ownership, and lifecycle contracts
- add provisional conformance traces plus C, C++, Rust, and Swift mapping notes
- keep every unresolved construct visible through `PROVISIONAL` markers and the RDL requirements ledger

## Status

Draft by design. No parser or generator consumes this syntax, and the PR does not change the HGL compiler or runtime build.

## Validation

- `git diff --cached --check`
- verified every `.hgspec` source begins with a `PROVISIONAL` notice
- documentation-only change; runtime suites are not required

PR #798, opened 2026-09-09
feat(hgl): compile initial standard library operators

hhenson:
## Summary

- compile hgraph.std as a real HGL standard-library module with len_ and is_empty implementations for strings, lists, sets, and maps
- install and export hgl::standard_library, including its generated registration descriptor, and exercise it from the installed-SDK consumer
- preserve source generic time-series parameters as TsVar during C++ schema emission while keeping scalar collection parameters as ScalarVar
- document the remaining production-operator, scheduling, rolling-window, and bundle-schema blockers explicitly

The generated code is clang-formatted, uses ordinary named static-node structs, and calls native free functions directly. It does not subclass operators. This slice intentionally proves the first two library migrations without replacing the production C++ operator registrations prematurely.

Stacked on #797. The backend-neutral hgspec prototype is tracked independently in #796.

## Validation

- focused HGL codegen and generated-runtime tests passed
- fresh native acceptance build: 1,801 tests passed
- Python 3.12 ABI3 wheel installed under Python 3.14: 3,323 passed, 10 skipped

Closes no issue.

### DIFF
--- a/language/stdlib/hgl/hgraph/standard.hgl
+++ b/language/stdlib/hgl/hgraph/standard.hgl
@@ -33,11 +33,7 @@ impl fn len_<T, const size: i64>(value: list<T, size>) -> i64 {
 
     when {
         let current = native::len(value)
-        if valid(out) {
-            if out != current {
-                out = current
-            }
-        } else {
+        if !valid(out) || out != current {
             out = current
         }
     }
@@ -48,11 +44,7 @@ impl fn len_<T>(value: set<T>) -> i64 {
 
     when {
         let current = native::len(value)
-        if valid(out) {
-            if out != current {
-                out = current
-            }
-        } else {
+        if !valid(out) || out != current {
             out = current
         }
     }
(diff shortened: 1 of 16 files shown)


language/stdlib/hgl/hgraph/standard.hgl
```diff
@@ -0,0 +1,129 @@
...
+        }
+    }
+}
+
+impl fn is_empty<T>(value: set<T>) -> bool {
+    inject out
+
+    when {
+        let current = native::is_empty(value)
+        if valid(out) {
```
hhenson: This could be better written as:
if !valid(out) || out != current { out = current }

Likewise with similar.

hhenson: Updated all six list, set, and map implementations to use the single guarded assignment in acdbff1ab. The generated code/runtime tests and full local acceptance gates pass.

Issue #810, opened 2026-09-09
Parity campaign: accept / fix / discuss decisions for 47 outstanding discrepancies


hhenson:
Every outstanding discrepancy found by the parity campaign, one item per finding,
each with **accept / fix / discuss**. Tick one box per item. Nothing here is
accepted yet: `known_divergences.json` is unchanged on all three open PRs, and no
fix issues have been filed.

* **accept** → I write the entry into `tools/parity/known_divergences.json` or
  `tools/parity/surface_known.json` and add a line to `parity_matrix.rst`, so the
  harness stops reporting it.
* **fix** → I open an issue and a PR against the candidate.
* **discuss** → it stays on this list and we talk it through.

Where I have a view, it is marked *(rec: …)*. Items marked **VERIFIED** I
reproduced by hand outside the harness, in both environments.

Sources: PR #804 (declaration-shape fuzzer), PR #806 (surface triage, 83 → 11),
PR #808 (operator coverage, 29 divergences + 5 from fuzzing).

---

## 1. Runtime and wiring defects

These change what a correct program does, or admit an incorrect one.

### 1.1 D28 — a graph's declared output type is not enforced **VERIFIED** *(rec: fix)*
`def g(ts: TS[bool]) -> TS[bool]: return invert_(ts)` returns `TS[int]`. The
reference raises a `WiringError`; the candidate accepts it and delivers
`[-2, -1, -2]` through a declared boolean output. A missing wiring check, not a
value difference, so it silently admits wrong programs. Plausibly the root of
the declared-vs-resolved family Alex has been patching at consumers.
- [ ] accept
- [x] fix
- [ ] discuss

### 1.2 D12 — `filter_` over a TSD loses state **VERIFIED** *(rec: fix)*
When the filter reopens, the reference republishes the whole accumulated map,
the candidate only the current delta. A consumer downstream never sees the
earlier keys. `gate`, `sample` and `if_` over the same TSD all agree, so it is
specific to `filter_`.
- [ ] accept
- [x] fix
- [ ] discuss

### 1.3 `log_` writes nothing on the candidate **VERIFIED** *(rec: fix)*
Released hgraph installs a stdout logging handler while the graph runs; this
runtime does not, so `log_`'s INFO records go nowhere, on stdout, stderr or at
file-descriptor level. `log_` is effectively a no-op. Secondary question if
fixed: does the handler belong to the runtime or to `GraphConfiguration`?
- [ ] accept
- [x] fix
- [ ] discuss

[...]

PR #844, opened 2026-09-10
Stop merge re-emitting when a removal leaves the value unchanged

### LINKED ISSUE #823: merge re-emits when a removal leaves the merged value unchanged
Decided **fix** on #810 item 7.2 (was N5 in PR #808 round 2). Found by the
random operator-family strategies.

`merge` re-emits when a key is removed from one input and the fallback input
holds the same value the removed key had. The reference emits nothing, because
the merged value has not changed.

Same reasoning as #810 item 6.3: the no-change-means-no-tick ruling
(2026-07-17, `roadmap.rst`) says an unchanged merged value should not tick, and
here it is the candidate that re-emits.

**Acceptance:** the N5 recipe matches the reference, and lands in the committed
corpus.

---

hhenson:
Closes #823. Decided **fix** on #810 item 7.2 (N5 in PR #808 round 2), found by
the random operator-family strategies.

## What was wrong

`merge` re-selects a source when the selected one goes away, and applied the
re-selected value **unconditionally**. So removing a key whose value the
fallback input already held re-emitted that value, even though the merged value
had not changed.

## The reference draws the line in exactly one place

`merge_ts_scalar` on `release/0.5` guards its re-selection branch:

```python
if tsl.modified:
    return next(iter(tsl.modified_values())).value
else:
    # This implies a value has gone away ...
    ...
    if out is not None and out != _output.value:
        return out
```

The **modified** branch has no such guard — an input that ticked is news
whatever it carries. The **fallback** branch does. This PR mirrors both halves:
the new `stdlib::apply_if_changed` sits beside `set_if_changed` in
`output_elision.h` and is applied to the three fallback sites in `merge_binary`
only.

This reaches `merge` over TSDs because both runtimes spell that
`map_(merge, *tsl)`, so a removed key lands in the per-key fallback — which is
the shape the campaign found.

## Verification

**A/B against the same build**, not against a stale artifact. The new C++ case
fails on the unfixed tree:

```
actual:   [..., {removed: {}, modified: {a: 1, b: 20}}]
expected: [..., {removed: {}, modified: {b: 20}}]
```

and passes with the fix.

**New parity recipe** `merge-tsd-removal-unchanged-no-retick` replays clean —
candidate and released 0.5.41 both trace `[{a:1,b:2}, null, {b:20}]`,
`difference: null`.

**The existing ported `test_merge_tsd` is untouched by the guard**, and that is
the point: its removal re-selects a *different* value, so it still ticks. The
new cases deliberately pair an unchanged fallback (`a`) with a changed one
(`b`) so both directions are pinned by one assertion.

Local gates: full C++ suite and full `python/tests` green; stdlib suite 1680
assertions / 445 cases.

## Doc

`roadmap.rst` — the no-change-means-no-tick ruling extended to cover a
**republication** as well as a projection, with the upstream guard quoted and
the fallback-only scope stated.

## Noted, not changed

The reference picks its fallback by *most recently ticked still-valid input*;
our binary fold picks structurally. The two coincide for two inputs and may
differ for three or more. That is a separate question from this issue and I have
not touched it.

---

### DIFF
--- a/docs/source/developer_guide/roadmap.rst
+++ b/docs/source/developer_guide/roadmap.rst
@@ -673,6 +673,21 @@ The following are intentional unless separately re-opened:
   which does **not** elide -- but that overload is not registered here at all,
   so there is no such path to diverge. If it is added, it needs its own
   implementation and must re-emit.
+
+  Extended 2026-09-10 (issue #823) to a **republication** -- an operator
+  re-publishing a value it did not compute. ``merge`` re-selects a source when
+  the selected one goes away, and if the re-selected value equals what merge
+  already published then nothing changed and nothing ticks. Released
+  ``merge_ts_scalar`` guards its own re-selection branch identically
+  (``if out is not None and out != _output.value``), so the ruling and upstream
+  parity agree here as well. The helper is ``stdlib::apply_if_changed``, beside
+  ``set_if_changed`` in the same header. It covers merge's **fallback path
+  only**: merge's modified path is unguarded in both runtimes, because an input
+  that ticked is news whatever value it carries.
+
+  This reaches ``merge`` over TSDs because both runtimes spell that
+  ``map_(merge, *tsl)``, so a removed key lands in the per-key fallback. Before
+  the fix, removing a key whose value the fallback already held re-emitted it.
 - **Reduce over partially-valid mapped keys** (issue #95; design record:
   :doc:`nested_graphs`): reduction is over currently-valid values. A keyed
   value can be invalid while its slot is live — a map child existing before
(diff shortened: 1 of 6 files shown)


hhenson:
Follow-up on the "Noted, not changed" section: I confirmed it rather than
leaving it as a suspicion, and filed it as #846.

The reference re-selects the **most recently ticked still-valid** input; our
static fold re-selects by **position**. For two inputs the rules coincide, so
nothing in this PR is affected. For three or more they diverge:

| tick 3 (the selected input is removed) | value |
|---|---|
| reference 0.5.41 | `{'k': 33}` (most recently ticked) |
| candidate | `{'k': 1}` (leftmost by fold position) |

It is independent of the elision this PR adds — the two candidate values differ,
so the guard does not fire either way — and it is unrecorded in both
`parity_matrix.rst` and `known_divergences.json`. Details and the repro are on
#846.

PR #903, opened 2026-09-11
sum_ over a window is a running sum, not a gated one

### LINKED ISSUE #857: [parity] stream_shape differs from released hgraph
## Differential result

A deterministic graph recipe passes with the maintained Python-first hgraph
0.5 reference and differs under C++-first hgraph. The case reproduced three times in fresh processes
and was reduced before this issue was created.

- Difference: `value` at `$.trace`
- Reference: `{'distribution': 'hgraph', 'machine': 'x86_64', 'platform': 'Linux-6.17.0-1022-azure-x86_64-with-glibc2.39', 'python': '3.14.7', 'python_implementation': 'CPython', 'version': '0.5.41'}`
- Candidate: `{'distribution': 'hgraph', 'machine': 'x86_64', 'platform': 'Linux-6.17.0-1022-azure-x86_64-with-glibc2.39', 'python': '3.14.7', 'python_implementation': 'CPython', 'version': '0.0.0'}`
- Original seed: `None`
- Origin case: `deviation-stream-to-window-min-period-ignored-upstream`
- Reduction: 2 accepted changes from 3 attempts
- Failure fingerprint: `463af60cb918eef294dad3f4133bf2d19e2611f7f1c8a03d44254c47bd472aab`

## Minimized recipe

```json
{
  "description": "Pin the accepted to_window deviation: released hgraph emits before min_window_period is satisfied, hgraph honours it. The divergence is upstream's. Accepted on issue #810 item 4.10.",
  "features": [
    "operation:to_window",
    "type:int",
    "family:stream",
    "deviation:min-window-period"
  ],
  "id": "deviation-stream-to-window-min-period-ignored-upstream",
  "inputs": {
    "ts": [
      0
    ]
  },
  "parameters": {
    "count": 2,
    "input_type": "int",
    "min_count": 2,
    "operation": "to_window",
    "reduction": "sum_"
  },
  "schema_version": 1,
  "template": "stream_shape"
}
```

## Expected reference trace

```json
{
  "canonical_schema_version": 1,
  "implementation": {
    "distribution": "hgraph",
    "machine": "x86_64",
    "platform": "Linux-6.17.0-1022-azure-x86_64-with-glibc2.39",
    "python": "3.14.7",
    "python_implementation": "CPython",
    "version": "0.5.41"
  },
  "phase": "complete",
  "process_returncode": 0,
  "recipe_fingerprint": "848175d3db46311905f084892751aa9cd567d7b84e01a623fa997b8f527e3941",
  "recipe_id": "deviation-stream-to-window-min-period-ignored-upstream",
  "schema_version": 1,
  "status": "ok",
  "trace": [
    0
  ]
}
```

## Actual C++-first hgraph trace

```json
{
  "canonical_schema_version": 1,
  "implementation": {
    "distribution": "hgraph",
    "machine": "x86_64",
    "platform": "Linux-6.17.0-1022-azure-x86_64-with-glibc2.39",
    "python": "3.14.7",
    "python_implementation": "CPython",
    "version": "0.0.0"
  },
  "phase": "complete",
  "process_returncode": 0,
  "recipe_fingerprint": "848175d3db46311905f084892751aa9cd567d7b84e01a623fa997b8f527e3941",
  "recipe_id": "deviation-stream-to-window-min-period-ignored-upstream",
  "schema_version": 1,
  "status": "ok",
  "trace": null
}
```

## Acceptance criteria

- The minimized recipe produces the released hgraph trace.
- The fix adds public Python wiring regression coverage.
- The same behavior receives equivalent native C++ `eval_node` coverage.

hhenson:
Closes #857.

## The bug

`tsw_numeric_aggregate_impl` serves both `sum_` and `mean` from one template, and gated **both** on `tsw_ready` — the minimum-window check. Upstream gates only the mean:

```python
sum_tsw    @compute_node(overloads=sum_)                        # no gate
mean_tsw   @compute_node(overloads=mean, all_valid=("ts",))
min_tsw    @compute_node(overloads=min_, all_valid=("ts",))
max_tsw    @compute_node(overloads=max_, all_valid=("ts",))
```

`sum_tsw` maintains a **running** total from `delta_value`/`removed_value`, so it answers from the first tick regardless of the minimum.

`sum_(to_window(ts, 2, min_window_period=2))`:

| ts | reference 0.5.41 | ours, before | ours, after |
|---|---|---|---|
| `[0]` | `[0]` | *(no ticks)* | `[0]` |
| `[5]` | `[5]` | *(no ticks)* | `[5]` |
| `[0, 1]` | `[0, 1]` | `[None, 1]` | `[0, 1]` |
| `[1, 2, 3]` | `[1, 3, 5]` | `[None, 3, 5]` | `[1, 3, 5]` |

## Why the gate is wrong for `sum_` only

Measured across the whole family at `min_window_period=2` (window 3, input `[1,2,3,4]`):

| operator | reference | ours, before | diverged? |
|---|---|---|---|
| `sum_` | `[1, 3, 6, 9]` | `[None, 3, 6, 9]` | **yes** |
| `mean` | `[None, 1.5, 2.0, 3.0]` | same | no |
| `min_` | `[None, 1.0, 1.0, 2.0]` | same | no |
| `max_` | `[None, 2.0, 3.0, 4.0]` | same | no |
| `std` | `[None, 0.5, 0.8164…, 0.8164…]` | same | no |

Only `sum_` diverged. That is what makes the gate wrong for `sum_` specifically, rather than the minimum-window notion wrong in general — an average over fewer points than were asked for is not the average that was asked for, but a running total over fewer points is still that total.

## The change

The gate alone. Partial contents were always readable — `tsw_ready` is a separate check (`window.size() >= window.min_period()`), not a bound on what the window view exposes — so `sum_` now needs only a non-empty window.

## Gates

- `ctest` — **1826/1826** (1824 baseline + 2 new).
- `uv run pytest python/tests` — **3486 passed**, 9 skipped, 0 failures (3484 baseline + 2 new).
- Both new Python tests pass **unmodified against released hgraph 0.5.41**.
- Verified by stashing the impl change: the `sum_` test fails without it (2 assertions); the `mean` guard passes either way, so the fix provably does not loosen the neighbours.
- Venv extension confirmed newer than the edited header before any Python result was trusted.

### DIFF
--- a/docs/source/developer_guide/parity_matrix.rst
+++ b/docs/source/developer_guide/parity_matrix.rst
@@ -39,7 +39,7 @@ Accepted deviations (decision list, 2026-09-09)
 
 The differential parity campaign (``tools/parity``) reported 47 outstanding
 discrepancies against released hgraph 0.5.41. Each was decided individually on
-issue #810 as *accept*, *fix* or *discuss*. The fifteen accepted here are
+issue #810 as *accept*, *fix* or *discuss*. The fourteen accepted here are
 permanent: released behaviour this runtime deliberately does not reproduce.
 Every one of them is either pinned by a fingerprint in
 ``tools/parity/known_divergences.json``, so the campaign exercises it and stops
@@ -71,9 +71,6 @@ Pinned by a corpus recipe and a fingerprint
      - ``set()``
      - ``{}``. The neighbouring ``str_`` renderings of a bool and a TSD are
        **not** accepted and are fixed under issue #819
-   * - ``to_window`` with ``min_window_period``
-     - Emits before the minimum period is satisfied, ignoring its own parameter
-     - Waits for the period. The divergence is the released implementation's
    * - Three-input ``intersection`` / ``symmetric_difference``
      - Fails at wiring: no set zero exists for the fold
      - Evaluates the fold. A superset, so no released program changes meaning
--- a/include/hgraph/lib/std/operators/impl/stream_impl.h
+++ b/include/hgraph/lib/std/operators/impl/stream_impl.h
@@ -315,7 +315,23 @@ namespace hgraph::stdlib
                 if (!ts.valid()) { return; }
                 const TSWInputView window_input{ts.base().borrowed_ref()};
                 auto window = window_input.data_view();
-                if (!tsw_ready(window)) { return; }
+                // ``mean`` waits for the minimum window -- an average over
+                // fewer points than were asked for is not the average that was
+                // asked for, and upstream's mean_tsw carries all_valid=("ts",)
+                // to say so. ``sum_`` does NOT: upstream's sum_tsw has no such
+                // gate and maintains a RUNNING total from the first tick, so
+                // it answers 1 for a window holding just [1] even when the
+                // minimum is 2 (parity #857). Partial contents are readable
+                // either way -- ``tsw_ready`` is a separate gate, not a bound
+                // on what the view exposes.
+                if constexpr (Mean)
+                {
+                    if (!tsw_ready(window)) { return; }
+                }
+                else
+                {
+                    if (window.size() == 0) { return; }
+                }
 
                 T total{};
                 for (std::size_t index = 0; index < window.size(); ++index)
--- a/python/tests/ported/_operators/test_tsw_operators.py
+++ b/python/tests/ported/_operators/test_tsw_operators.py
@@ -15,6 +15,37 @@ def test_tsw_sum():
     assert eval_node(g, [1, 2, 3]) == [1, 3, 6]
 
 
+def test_tsw_sum_is_a_running_sum_below_the_minimum_window():
+    """``sum_`` over a window does NOT wait for the minimum window period.
+
+    Upstream's ``sum_tsw`` carries no ``all_valid`` gate and maintains a
+    running total from the first tick, so it answers 1 for a window holding
+    just [1] even when the minimum is 2. Its neighbours ``mean``/``min_``/
+    ``max_`` DO carry ``all_valid=("ts",)`` -- an average over fewer points
+    than were asked for is not the average that was asked for -- which is why
+    only ``sum_`` behaves this way (parity #857).
+    """
+
+    @graph
+    def g(ts: TS[int]) -> TS[int]:
+        return sum_(to_window(ts, 2, min_window_period=2))
+
+    # A single tick, below the minimum of 2: still summed.
+    assert eval_node(g, [0]) == [0]
+    assert eval_node(g, [5]) == [5]
+    assert eval_node(g, [1, 2, 3]) == [1, 3, 5]
+
+
+def test_tsw_mean_still_waits_for_the_minimum_window():
+    """The guard for the above: the gate is dropped for ``sum_`` only."""
+
+    @graph
+    def g(ts: TS[int]) -> TS[float]:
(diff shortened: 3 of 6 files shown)


hhenson:
## ⚠️ This PR retires an accepted deviation — and this one is a judgement call

Same discovery as on #900: comparing parity-smoke counts across my open PRs shows a recipe moving from *known mismatch* to *matched*.

| PR | matched | known mismatches |
|---|---|---|
| #898, #901, #904, #905 (neutral) | 279 | 35 |
| **#903 (this one)** | **280** | **34** |

Retiring an accepted deviation **passes CI silently** — only changing the *shape* of one fails the gate. So this had to be looked for.

The recipe is `deviation-stream-to-window-min-period-ignored-upstream`, accepted as **#810 item 4.10**:

> "released hgraph emits a to_window reduction before min_window_period is satisfied, so it ticks at the first element where hgraph waits. **The divergence is the released implementation ignoring its own parameter.**"

## Why I think the premise is wrong

Upstream does not ignore the parameter uniformly. In `_tsw_operators.py`:

```python
sum_tsw    @compute_node(overloads=sum_)                        # no gate
mean_tsw   @compute_node(overloads=mean, all_valid=("ts",))
min_tsw    @compute_node(overloads=min_, all_valid=("ts",))
max_tsw    @compute_node(overloads=max_, all_valid=("ts",))
```

Three of four gate; the one that doesn't keeps a **running** total from `delta_value`/`removed_value` and never reads `ts.value` — the property that returns `None` below the minimum. It could not honour the minimum without being a different algorithm.

And measured at `min_window_period=2`, **only `sum_` diverged** — `mean`, `min_`, `max_` and `std` already agreed with us that they wait. A parameter ignored by oversight would not be applied to three operators out of four and skipped on precisely the one where it cannot apply.

## But unlike #900, this is a reading, not a refutation

On #900 the record claims "the output TSD is empty" and `len_` is measurably `1` — the premise is just false. Here the record claims upstream is ignoring its own parameter, and I'm arguing from the *shape* of upstream's code that it's deliberate. That's interpretation. Someone decided "accept" once already, and reversing it is your call, not mine.

There's a defensible case for the original reading too: the TSW type defines `all_valid` as `_length >= _min_size`, so the notion exists at the type level and `sum_tsw` simply doesn't opt in. Whether that's intent or oversight isn't provable from the source.

## What I've done

Removed the now-dead entry and re-described the corpus recipe to pin the agreement, so merging as-is doesn't leave a record claiming a divergence that no longer exists. **The last commit reverts cleanly** if you'd rather keep item 4.10 as accepted — that would also reopen #857, which was filed against this shape.

hhenson:
**Merge note:** entries `c0525119` (#810 item 4.10) and `c3ffb967` (#810 item 6.1) are **adjacent** in `known_divergences.json`, and #900 and #903 each remove one of the pair. Git cannot combine the two deletions, so whichever merges second conflicts on that file.

The correct resolution is to **delete the whole `{...}` object** — each side removed one of the adjacent pair, so neither should remain. A careless resolution restores a dead entry.

I verified the resolved result: 10 divergences and 11 families, with `c0525119`, `c3ffb967` and `9d3ebf73` (the rshift entry retired by #902) removed, and `a3d2bb7d` (lshift 4.7), `aecb307b` (empty-set 4.4) and `7834fe8e` (mesh no-tick) all still present. All seven PRs then merge cleanly: **ctest 1835/1835, pytest 3504**, zero failures.

PR #921, opened 2026-09-13
feat(hgl): add const functions and temporal lifting

hhenson:
## Summary

- Add module-local `const fn` value functions, distinct from parameter-level `const` and temporal `fn` declarations.
- Apply temporal-first selection in graph composition and `eval`: an existing temporal definition wins, and a signature error never triggers fallback. `const(function)` explicitly selects the value role.
- Lower temporal calls to value functions through shared default-policy runtime adapters used by both AOT C++ and scripted execution. All-scalar calls execute directly; native phase restrictions propagate through value helpers.
- Emit plain C++ value helpers once in dependency order, diagnose recursion, exercise lifting in the native string library's HGL tests, and refresh generated C++ snapshots.
- Document the selection rules, examples, implementation layers, and remaining boundaries.

## Stack

Depends on #920; this PR targets `codex/hgl-native-parts-coverage` to isolate the lifting changes. Merge the parent first and retarget this PR to `main`.

## Validation

- macOS: fresh native acceptance configuration with HGL enabled, clean rebuild, and complete CTest suite — **1,858 passed**.
- Private Linux validation host: GCC 14 native build and complete CTest suite — **1,858 passed**, including the installed-SDK consumer with dependency packages configured.
- Both platforms: stable-ABI wheel built with Python 3.12, installed into a fresh Python 3.14 environment, and complete non-WIP compatibility suite — **3,466 passed, 10 skipped** each.
- Executable HGL lifting fixture: eight tests, plus matching public C++ wiring coverage and semantic rejection tests. Forward dependencies exercise both concise and block bodies; direct and mutual recursion and unsupported structural value signatures have rejection coverage.
- Documentation relative links and `git diff --check` passed.

## Deliberate boundaries

This slice supports local, non-generic, fixed-arity value functions with scalar parameter/result types (or a void result). Structural value parameters/results are rejected during checking until their runtime-value and ownership conversions are implemented; schema markers must not be emitted as payload types. This includes explicitly `const` structural parameters. Generic/pack value functions are explicitly diagnosed. Exported/imported value descriptors, `impl`/`native` modifier combinations, native-family selection, and general first-class callable storage remain follow-up work. Existing value-expression backend limitations are documented rather than presented as implemented.

Remote CI is not included in the local validation results above.


language/docs/user-guide/value-functions.md
```diff
@@ -0,0 +1,119 @@
...
+| `when`, `start`, `stop`, or another `const fn` | Select the value `const fn`; temporal calls are unavailable |
+| `eval` | Select the temporal `fn` if declared; otherwise test the lifted `const fn` |
+
+`const(scale)` explicitly selects the value declaration, overriding the first
+and third rules. It does not itself construct a node or call the function:
+
+```hgl
+fn scale(value: f64, const factor: f64 = 2.0) -> f64 {
+    when {
+        return const(scale)(value, factor) + 10.0
```
hhenson: Inside a when clause cont selection is implicate and does not require qualification.

hhenson: You could put a comment in to explain the example’s selection

hhenson: Updated in 347221760. The when example now calls scale(value, factor) directly, with an inline # comment explaining that the const fn is selected implicitly in this context. The explicit const(scale) selector remains in the graph example where it overrides normal temporal-first selection. The executable HGL fixture covers this implicit selection.

language/src/codegen/cpp_emitter.cpp
```diff
@@ -5183,6 +5257,7 @@ namespace hgl::codegen
             std::map<std::uint32_t, std::set<std::uint32_t>> deps;
             for (const gir::CallableId id : internal) {
                 const gir::Callable &fn = callable(id);
+                if (fn.kind == gir::CallableKind::ValueFunction) { continue; }
```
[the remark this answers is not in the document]
hhenson: Fixed in 347221760. Value functions now participate in dependency collection and topological ordering, including concise and block bodies, so later-declared callees are emitted first and direct/mutual recursion reaches the existing diagnostic. Helpers are emitted once in the header; private adapters and exported inline node hooks share that definition. Added ordering/recursion regression tests and executable HGL plus public C++ coverage for forward calls through both lifted and explicit runtime nodes. All 1,858 native tests passed on macOS and Linux; the eight-test HGL fixture passed on both.

language/src/codegen/cpp_emitter.cpp
```diff
@@ -4980,6 +5013,47 @@ namespace hgl::codegen
...
+                frame.fn               = decl;
+                frame.runtime          = true;
+                frame.output_available = true;
+                std::vector<std::string> parameters;
+                for (const gir::Parameter &parameter : planned.parameters) {
+                    const HType       type = planned_type(parameter.type, planned.range);
+                    const std::string name = cpp_name(parameter.name);
+                    const std::string type_name = value_type(type, planned.range);
+                    const bool borrow = type.kind != HType::Kind::Scalar || type.is(hir::ScalarType::Str);
+                    parameters.push_back("[[maybe_unused]] " + (borrow ? "const " + type_name + " &" : type_name + " ") + name);
```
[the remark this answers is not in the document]
hhenson: Fixed by rejecting unsupported non-scalar const fn signatures during HIR checking in 347221760, before either backend can emit schema markers as runtime payloads. This covers parameters and results, including explicitly const parameters, with regression cases for sets, maps, lists, tuples, structs, and atomic structural types. Scalar signatures remain supported. The guide, ADR, roadmap, and PR description explicitly record structural runtime-value/ownership lowering as follow-up work; this change does not claim to implement that ABI. Both complete native suites (1,858 tests each) and Python compatibility suites (3,466 passed each) are green.

hhenson:
Merged the #920 review fixes into this branch in 6a8fd59d9, preserving the const-function/lifting changes and their previous review fixes. The combined stack passes a fresh, clean macOS language-enabled native build and all 1,858 CTest tests, including generated C++ snapshots and installed consumers. A newly built Python 3.12 stable-ABI wheel passes the full Python 3.14 non-WIP suite: 3,466 passed, 10 skipped. Parent #920 additionally passed full native and Python acceptance on a private Linux validation host and 100 HTTP/2 loopback repetitions on each platform. Remote CI is rerunning separately.

PR #955, opened 2026-09-16
Prototype: dmap_ experiments validating RFC 0037

hhenson:
**Prototype only — not for merge into the runtime.** Everything lives under
`prototypes/`, which is outside `testpaths = ["python/tests"]` and
`wheel.packages = ["python/hgraph"]`, is in no CMakeLists, and is imported by
nothing in the main tree. Deletable with one `rm -rf`.

Validates the v0 model in [RFC 0037](https://github.com/hhenson/hgraph/pull/954).

## Result

All five experiments produce output **identical to `map_`** at 1, 2 and 3
workers, in genuinely separate OS processes (verified by pid).

| | what it would catch |
| --- | --- |
| E1 stateless | partition routing, delta round trip, output merge |
| E2 stateful sum | per-key state landing in the wrong worker |
| E3 tick count | a dropped/duplicated dispatch (output is pure tick history) |
| E4 shared input | a broadcast value not reaching every worker with keys |
| E5 key lifecycle | a child built or torn down in the wrong worker |

**RFC 0037's central bet paid**: the worker hosts an *ordinary* `map_`, so no
per-key lifecycle code had to be written — E5 passes on existing behaviour.

## The cost (128 cores, raw JSON committed)

Ratio is `dmap_ / map_`; lower is better.

| keys | per-key work | 1 | 2 | 4 | 8 | 16 |
| --- | --- | --- | --- | --- | --- | --- |
| 64 | trivial | 14.2x | 14.5x | 14.4x | 14.9x | 16.8x |
| 64 | ~60us | 0.88 | 0.56 | 0.31 | 0.18 | **0.14** |
| 256 | trivial | 5.2x | 5.0x | 4.9x | 5.1x | 5.7x |
| 256 | ~60us | 1.00 | 0.51 | 0.26 | 0.15 | **0.12** |

Best **8.3x faster**; worst **37x slower**. Trivial children are hopeless and
more keys does *not* rescue them — the overhead is per cycle and per worker.
One worker is never worth it.

## Known gap, stated rather than papered over

Engine time is **not** external. `GraphView::evaluate` is not exposed to
Python, so the worker runs `REAL_TIME` on its own clock. One dispatch is still
one engine cycle, so the *sequence* is lockstep — but `next_scheduled_time`
propagation and any self-scheduling child are **entirely unexercised**. That is
half the RFC's per-cycle contract. Every v0 kernel is time-independent, which
is what makes these results sound and also what bounds them.

## hgraph behaviours found the hard way

- An **unwired optional `TS[...]` input stops a node from ever evaluating** —
  silently, empty result, no error. Needs `@compute_node(valid=(...))`.
- **`@generator` is pulled with one item of lookahead**, so it cannot drive a
  lockstep worker (it would block on cycle T+1 before evaluating T). Hence
  `push_queue`.
- **`hash()` cannot partition** — CPython salts string hashing per process.


hhenson:
Prototype succeeded by impl

PR #957, opened 2026-09-16
dmap_: a map_ whose children run in worker processes (RFC 0037)

hhenson:
`dmap_` evaluates mapped child graphs in worker processes using the ordinary native `map_` wiring and lifecycle. It supports dictionary and fixed/dynamic list mappings, multiple and named inputs, broadcasts, `pass_through`, `no_key`, explicit keys, custom or disabled key injection, scalar configuration, and outputless children. Both C++ and Python clients use the same prepared plans and worker pool.

This PR includes RFC 0037, the externally driven executor, native process boundaries, binary codecs, framed transport, registered C++ worker recipes, and native performance measurements. It incorporates `main` through `f67b050f0` and addresses the four original review findings.

## Boundary semantics

- Recursively transfer `TS`, `SIGNAL`, `TSS`, `TSD`, fixed/dynamic `TSL`, `TSB`, and `TSW`. Materialize references at process boundaries while preserving their behavior inside a child.
- Preserve invalid dictionary membership, partial structures, list holes and shrink/regrowth, removal/recreation, final-child invalidation, and original window timestamps.
- Transfer ordinary window samples/clears incrementally; send full history on initialization or rebinding. Capture and encode common input changes once before fan-out.
- Support portable numeric, text, bytes, temporal, enum, compound and collection scalars, Arrow frames/series, and owned/shared payloads. Bound converters retain immutable type realizations and closed polymorphic alternatives. Zone identities and partition hashes are process-independent.
- Partition list children by original ordinal. Each worker allocates and evaluates its own children; broadcast input storage and transport still scale with worker count.

## Python and process constraints

Importable module-level Python graphs/nodes and registered native operators are reconstructed in fresh interpreters. Portable scalar configuration and type recipes cross the bootstrap boundary; executable code, closures, opaque Python objects, and live handles do not.

Parent services, contexts, captured ports, and push sources are rejected before launching workers. Worker configuration and runtime `GlobalState` are isolated; transport uses reserved keys. Python wiring-state views expire when composition returns. Worker restart, rebalancing, and component checkpoint recovery remain outside this implementation. Worker/sink failures propagate, and side effects across partitions are not transactional.

The prepared API replaces the former single-scalar-input limitation. The legacy scalar-only C++ recipe API remains available with its documented canonical-delta limitations; new clients should use prepared plans.

## Review hardening

Worker process sends and replies share a configurable finite deadline (60 seconds by default), and failures terminate the pool. POSIX sockets suppress SIGPIPE and use nonblocking descriptors; Windows uses cancellable overlapped pipes and UTF-16 process creation. Python bootstrap metadata crosses the connected channel, avoiding command-line size limits.

Frames are capped at 64 MiB. Decoders enforce shared work/depth budgets, slot direction and uniqueness, and malformed-prefix checks. Boundary deltas are decoded once and validated before live mutations, including sparse list extents, duplicate children, and window history chronology. Default and moved-from binary converters fail predictably. Documentation explicitly limits workers to trusted code; process separation is not a security sandbox.

## Validation

The distributed Python corpus contains 264 tests, including ordinary `map_` oracles, in-process execution, and real process workers. Native coverage includes matching public wiring scenarios, process recipes, invalid membership and slot reuse, partition lifecycle, temporal/Arrow/polymorphic codecs, and constant-sized window warmup payloads. Bound converter tests verify that steady-state encode/decode avoids type-system locks.

Stable-ABI wheels are built with Python 3.12 and tested in fresh Python 3.14 environments.

| Platform | Native suite | Python compatibility | Persistence compatibility |
|---|---:|---:|---:|
| macOS | 2,050 passed | 3,751 passed, 10 skipped | 642 passed, 1 skipped |
| Private Linux host, GCC 14.3 | 2,050 passed | 3,751 passed, 10 skipped | 643 passed |
| Private Windows host, MSVC 19.51 | 2,049 passed | 3,751 passed, 10 skipped | 642 passed, 1 skipped |

Linux ASan passed all four core test binaries (1,945 cases) plus 14 checkpoint-store cases; leak detection was disabled. Core and persistence installed-SDK consumers pass on all three platforms. Sphinx builds with warnings treated as errors. The TSData operations ABI advances to 20 for structural membership and timestamp-preserving window restoration; the persistence extension is rebuilt and tested against it.

A focused macOS dense-list decode/apply check (50 full deltas per size) measured 93.0 / 175.3 / 353.9 / 701.4 microseconds for 1,000 / 2,000 / 4,000 / 8,000 children, respectively. This supports linear scaling for that validation path; it is not an end-to-end worker throughput benchmark.

The existing native benchmark corpus measures workload-dependent transport crossover, not Python startup/callback costs. Incremental transfer is proportional to changed data times worker count. GitHub CI was not monitored, as requested.


hhenson:
@codex review

Please perform a fresh adversarial review of the current head, 23dbb96e621e44ddf39c4361923c4734e78e9740. The previous review predates the expanded distributed-map implementation.

Focus on map_ behavioral parity across supported call shapes and recursive time-series types, C++/Python parity, reference materialization, invalid collection membership, window history and incremental transfer, worker state isolation, lifecycle and failure cleanup, and algorithmic complexity. Look for missing regression coverage and architectural or readability problems. Reassess the current implementation rather than relying on resolved findings from the earlier revision.

hhenson:
@codex security review

Please review the current head, 23dbb96e621e44ddf39c4361923c4734e78e9740. The implementation has expanded substantially since the previous security review.

Focus on binary decoding and resource bounds, malformed process messages, worker bootstrap and command-line handling across platforms, lifetime and ownership safety, state isolation, and cleanup after partial startup or worker failure. Assess the documented trusted-code process model without assuming that worker processes provide a security sandbox.

PR #1001, opened 2026-09-19
Support reconstructible cache beside recordable node state

hhenson:
Native static nodes previously rejected `State<T>` beside `RecordableState<TSchema>`, preventing durable semantic state from having a reconstructible local cache. Allow one of each using the existing independent storage slots. Checkpoint recovery restores durable state before `start` rebuilds the fresh cache; ordinary-state-only compute nodes and unsupported runtime services remain refused.

Python compute nodes now expose the same native lifecycle through `STATE`/`STATE[T]` plus `RECORDABLE_STATE`. Typed cache class identity participates in checkpoint compatibility, and local state is released after normal execution, lifecycle failures, and execution without a stop hook. Regression coverage includes repeated restores, hidden durable updates without public output, compute and sink nodes, cache destruction, and an installed-SDK mixed-state consumer.

Update the authoring contract, skill guidance, HGL roadmap and diagnostics. HGL mixed state/cache lowering remains explicitly gated; recoverable pending scheduling is separate follow-up work. Cache reconstruction remains the node author's responsibility.

Acceptance also exposed existing validation issues: worker teardown duplicated the shared exception recorder, a worker test bypassed the portable environment helper, and the Windows SDK runner omitted the base Python DLL directory. Use the existing helpers and honour configured build parallelism.

Validation:
- macOS and private Linux validation host: fresh native acceptance configuration and clean build, then all 2,272 CTest tests passed on each platform, including HGL integration tests.
- Python 3.12 stable-ABI wheels installed into fresh Python 3.14 environments: 3,908 non-WIP compatibility tests passed and 10 skipped on each platform.
- Persistence compatibility: macOS 690 passed/1 skipped; Linux 691 passed.
- Private Windows validation host, MSVC 19.51: 2,182 native tests passed; stable-ABI wheel built and all nine new Python lifecycle tests passed.
- Installed-wheel SDK consumers passed on macOS, Linux and Windows.
- Linux AddressSanitizer with leak detection: all four new native test cases passed in separate processes (58 assertions), avoiding the documented registry-reset retention between test cases.
- Sphinx dummy build, HGL catalogue consistency and whitespace checks passed.

Remote CI is separate from these local validation results.


hhenson:
Merge conflict with `main` repaired in `c43d96f47`. Two conflicts, both against the recovery review follow-ups (#998):

- **`include/hgraph/runtime/distributed_map.h`** -- both sides had made the *same* fix to `WorkerPool::stop` (`FirstExceptionRecorder` in place of a bare `catch (...)`, for the `catch-all-sites` ratchet), differing only in a variable name. `main`'s wording is kept, so this branch carries no diff there.
- **`python/py_nodes.cpp`** -- a real merge, and both changes are kept. This branch signs a typed `STATE` cache by its class identity instead of its value; `main` turns an unsignable scalar into a named refusal rather than a bare `checked_as<T> type mismatch`. The cache branch is untouched, and the ordinary-scalar branch takes `main`'s annotated signing.

`docs/source/user_guide/component_recovery.rst` was the only other file changed on both sides; it merged by itself, and the two sets of additions are in separate sections and agree (a node with *only* local state is still refused, which this PR says too).

Validated at `c43d96f47` on a private Linux validation host:

| What | Result |
|---|---|
| Full C++ build, GCC 14 `-Werror`, full `ctest` | 2,192 / 2,192 |
| **Real install** (core wheel, then the persistence extension against its SDK): `python/tests` **and** `extensions/persistence/python/tests` | 4,612 passed, 9 skipped, 1 xfailed, 0 failed |
| Architecture ratchets on the merged tree | 21 / 21 |

The xfail is the strict pin of the key-order defect from #999, not this PR's.

One note for whoever next uses the branch's existing local worktree: it had nothing unpushed, but it is now two merges behind the remote and needs a pull.

PR #1013, opened 2026-09-19
Fix restored keyed inputs changing key order: slot-store growth now commutes with the erase flush

hhenson:
## What was wrong

A restored keyed input iterated its keys in another order from the unbroken run — `(3, 0)` where the unbroken run gave `(0, 3)` — and an order-sensitive `reduce` showed it as `40` against `76`. The recovery campaign found it and filed it as the known-defect family `tsd-restored-slot-order`, with the root cause unknown and a strict `xfail` pinning the minimal stream.

## Root cause

A checkpoint **normalizes** pending erases: a keyed image records the free pool as it will stand *after the next flush* (`KeySlotStore::checkpoint_free_slots`). A restored store has therefore already flushed where the uninterrupted one has not yet. That is sound only if everything else that touches the free pool commutes with the flush.

* Allocation does: the owner flushes at the first mutation of a new evaluation time, before any insert.
* **Capacity growth did not.** `reserve_to` pushed new slots on *top* of the pool, and the Python result path reserves room for its result *before* its first mutation (`apply_dict_result`, `mutation->reserve(source.size())`).

With a delta larger than the capacity, on the minimal stream `[{2:2}, {2:REMOVE, 1:4}, {0:7, 1:REMOVE, 3:3}]` cut at 2:

| | pool before the inserts | key `0` | key `3` |
|---|---|---|---|
| uninterrupted: grow, then flush | `[2, 0]` | slot 0 | slot 2 |
| restored: slot already free, then grow | `[0, 2]` | slot 2 | slot 0 |

Keys iterate in slot order, hence `(3, 0)`.

This also answers the two things that were unexplained. The C++ counterpart passed on the same stream because the C++ replay source applies a delta without ever calling `reserve`. And "a removal on each side of the cut" was the relation only because the second removal made the delta three entries against a capacity of two.

## The fix

`KeySlotStore::reserve_to` adds new capacity **underneath** the free pool instead of on top. Holes are reused before fresh slots, and the pool is identical whichever of growth and the flush came first. The rule lives in the store that owns the pool rather than in each caller that might reserve early. It is linear in the pool and paid only when capacity actually grows (`acquire_free_slot` only grows an empty pool, so that path is unchanged).

## Tests — each fails before the fix

* `test_slot_utils.cpp`, *"KeySlotStore capacity growth commutes with the pending-erase flush"*: either order leaves one free order (`{4,3,2,0}` vs `{0,4,3,2}` before); a restored store allocates as the uninterrupted one does.
* `test_ts_checkpoint.cpp`, *"a dictionary reserved ahead of its first mutation keeps its key order"*: the campaign's minimal stream through capture and restore, with the reserve a Python node makes — a pure C++ reproduction of what only Python showed.
* `test_reduce_recovery_scenarios.py`: the strict `xfail` comes off; the minimal stream is an ordinary test.
* The campaign's family and its key-order probe are deleted; `test_recovery_campaign.py` pins the minimal stream as belonging to **no** family, so a recurrence is a plain failure.

## Docs (same change)

* RFC 0023, "Dynamic nested graphs": records that the implementation normalizes deferred erase state and the commutation rule that makes that sound; the open question on exact pending-erase capture is resolved.
* `plans_and_ops/time_series.rst`: the free-pool rule on `KeySlotStore`.
* `testing.rst`: one known family remains (`mesh-empty-input-no-tick`, the consequence of a ruling, not a defect).

## Validation

* macOS, full native suite: 2,182 cases — 2,181 passed, 1 failed as expected (the pinned `dmap_` contract limit, unrelated).
* Architecture ratchets: 21 passed.
* GCC 14 `-Werror` + full ctest, the persistence Python suite on a real install, and the recovery campaign on nightly seed 2 (the seed that showed six failures in the family) are running on a private Linux validation host; results follow in a comment.


hhenson:
## Validation

The behaviour was validated at `a0ac4e007`; the head `22f79a970` adds a comment and an RFC paragraph only (no code).

**macOS arm64 (clang)**
* Native suite: 2,182 cases — 2,181 passed, 1 failed as expected (the pinned `dmap_` contract limit); ctest 2,216 / 2,216.
* Python, on a real install of this branch with the persistence extension built against its SDK: persistence suite 705 passed, 1 skipped; core suite 3,915 passed, 9 skipped.
* Sphinx `-W` HTML build clean; architecture ratchets 21 passed.

**Linux x86_64 (GCC 14, `-Werror`), a private validation host**
* Native build clean, ctest 2,203 / 2,203.
* Persistence Python suite on a real install: 706 passed.
* **Recovery campaign, nightly profile, seed 2** — the seed that showed six failures in the retired family: 3,480 of 3,480 scenarios pass (`recover` 424/424, `snapshot` 3,056/3,056), none filed as known, controls 694/724 sensitive, verdict exit 0.

**Two things the first push got wrong, both corrected in the history above**
* GCC 14 failed `test_slot_utils.cpp` with `-Werror=array-bounds`. I first blamed the `std::rotate` this PR had added and replaced it; the diagnostic was actually in an untouched test 800 lines away, exposed because adding a test case changed what GCC inlines in that file. Both stable slot store implementations re-read `slot_count_` after a call and before a copy, so GCC could no longer see the count is below the capacity. They now read it once (`a0ac4e007`). The rotate replacement stays on its own merit — the grown pool is built aside, so a failed allocation leaves the store as it was — and the commit message says the first diagnosis was wrong.
* A comment claimed the store guarantees allocation never precedes the flush. It does not; that is the owner's discipline (`TSS`/`TSD` in `prepare_delta`, `mesh_` in `erase_retired_before`). Reviewing that also showed `mesh_` was exposed to the same defect by another route — it mirrors the capacity of its requested key set, which grows before the mesh evaluates and flushes. The store-level fix covers it; RFC 0023 now records both (`22f79a970`).

